PR #19, opened 2022-05-19
WIP update for LWJGL 3.3.0

unit-a-user:
Indigo renderer issue: https://github.com/orgs/FabricMC/discussions/2231 (discussion)
It's works when "you have to set always-tesselate-blocks=false in config/fabric/indigo-renderer.properties"

What I changed: 
- ~~replaced `com.example` to `net.vulkanmod`~~
- ~~update `yarn_mappings=1.18.2+build.3`~~
- ~~update `fabric_version=0.51.1+1.18.2`~~
- ~~update `loader_version=0.14.5`~~
- ~~increased memory `org.gradle.jvmargs=-Xmx4G`~~
- tried `project.ext.lwjglVersion = "3.3.0"`
- fix `allocatorCreateInfo.instance(instance);` in `Vulkan.java`
- ~~avoid `.vs` or `.vc` folder from upload~~
- rise back byte color instead of float, back compatibility with indigo tessellation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,8 +31,3 @@ bin/
 # fabric
 
 run/
-
-# avoid bad open
-
-.vs/
-.vc/
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,8 @@ bin/
 # fabric
 
 run/
+
+# avoid bad open
+
+.vs/
+.vc/
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ archivesBaseName = project.archives_base_name
 version = project.mod_version
 group = project.maven_group
 
-project.ext.lwjglVersion = "3.2.3"
+project.ext.lwjglVersion = "3.3.0"
 project.ext.jomlVersion = "1.10.4"
 project.ext.winNatives = "natives-windows"
 project.ext.linuxNatives = "natives-linux"
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ archivesBaseName = project.archives_base_name
 version = project.mod_version
 group = project.maven_group
 
-project.ext.lwjglVersion = "3.3.0"
+project.ext.lwjglVersion = "3.3.0" // needs to be save 3.2.3?!
 project.ext.jomlVersion = "1.10.4"
 project.ext.winNatives = "natives-windows"
 project.ext.linuxNatives = "natives-linux"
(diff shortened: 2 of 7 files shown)


unit-a-user:
It's works when "you have to set `always-tesselate-blocks=false` in `config/fabric/indigo-renderer.properties`"

arlomcwalter:
why are you allocating 8gb to a development environment? seems very excessive

unit-a-user:
First of all, it is not a fact that 8Gb, because the flag indicates the maximum capacity. Secondly, just in case, in case an overflow or slowdown occurs.

polarbub:
How do you get around Minecraft trying to load its own lwjgl?

polarbub:
For blaze4d I use multimc to disable the default lwjgl 3.2.3 and use 3.3.1

arlomcwalter:
> First of all, it is not a fact that 8Gb, because the flag indicates the maximum capacity. Secondly, just in case, in case an overflow or slowdown occurs. Do you propose to leave it like that, or to assign only 4Gb? Thirdly, systems now rarely have less than 8 or 16 Gb of RAM, with rare exceptions.

i understand, lol. it's just very excessive, i've never even had an issue with capping it at 2gb. sure it wont necessarily use 8gb but thats not a reason to max it at 8gb, what if someone only has an 8gb system? (not very uncommon)

polarbub:
Does it work in the vanilla launcher?

unit-a-user:
Mod probably yes, IDK.
But update LWJGL for vanilla probably not possible due recursive hashing.

polarbub:
Are you saying you used the customize button in multimc for lwjgl?

unit-a-user:
> Are you saying you used the customize button in multimc for lwjgl?
Yes, but I'm also used JavaScript script, because I'm tired writing every checksum by hand. 
Also, final answer about official launcher: **YES!**

unit-a-user:
Also, `R8G8B8A8_UNORM` is works. And `always-tesselate-blocks=auto` is works!

xCollateral (changes requested):
Would you remove non reviewed files from pr?
Besides that good work.

unit-a-user:
> Would you remove non reviewed files from pr? Besides that good work.

I apologize for the question of what files are we talking about, specify?

xCollateral:
Gradle settings files, and `Vulkan.java`. `.gitignore` is fine.
Let every user chose these config.

unit-a-user:
Rise back `gradle.properties` and `.gitignore`.

xCollateral:
I'll merge this, but i think i'll just update to lwjgl 3.31 later.

xCollateral (approved):

PR #23, opened 2022-06-08
[WIP?] Add support on macOS

QianNangong:
It does work well on macOS with Apple M1, but I'm not sure if it also works on Intel Macs with Intel or AMD graphic cards... 
image


xCollateral:
Thank you for your test. I didn't want to include macos native libs due to increase in jar size that is already heavy.
I'll close this pr until i decide what to do.

PR #35, opened 2022-06-16
Basic frustum changes

tr7zw:
- Use a parallelStream to recalculate the frustum, so the cpu is better utilized
- Add a bit? of overdraw, but heavily reduce how often the frustum is recalculated

This slightly reduces the fps while standing still, but can up to double the fps while moving the camera/player.
In the future this should probably work more like Sodium does: with a flood fill and better optimized Frustums.

### DIFF
--- a/src/main/java/net/vulkanmod/mixin/WorldRendererMixin.java
+++ b/src/main/java/net/vulkanmod/mixin/WorldRendererMixin.java
@@ -1,9 +1,12 @@
 package net.vulkanmod.mixin;
 
 import com.mojang.blaze3d.systems.RenderSystem;
+
 import it.unimi.dsi.fastutil.objects.ObjectArrayList;
 import it.unimi.dsi.fastutil.objects.ObjectListIterator;
 import net.minecraft.client.render.*;
+import net.minecraft.client.render.WorldRenderer.ChunkInfo;
+import net.minecraft.client.render.WorldRenderer.class_6600;
 import net.vulkanmod.interfaces.ShaderMixed;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gl.ShaderEffect;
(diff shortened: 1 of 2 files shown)


andreasdc:
> * Use a parallelStream to recalculate the frustum, so the cpu is better utilized
> * Add a bit? of overdraw, but heavily reduce how often the frustum is recalculated
> 
> This slightly reduces the fps while standing still, but can up to double the fps while moving the camera/player. In the future this should probably work more like Sodium does: with a flood fill and better optimized Frustums.

Nice to see such movement on this mod. :) I was testing it for couple minutes, but I don't know how to check if it helped.

tr7zw:
To test the difference: Stand still and circle the mouse. The fps will drop by a lot, with this change a bit less(but still drop).

SpidFightFR:
@tr7zw i tested your build, indeed, we loose less FPS with your change (i noticed a maximum loss of 20 fps: from 100 to 80fps). 
Good job, mate ! 👌

andreasdc:
Will it help for this? Maybe it can be implemented?
image

tr7zw:
Yea it's exactly what this pr aimed to improve, but it's janky(and outdated). Probably better to fix it correctly like Sodium/Optifine do.

andreasdc:
Would be nice to see the fix 🚀

xCollateral:
Latest update addresses this aspect in a better way, from how chunks are updated to optimized frustum culling. Frustum computations could still be made multi threaded but the benefits would be minimal. I'll close this PR for now, if you still want to try or come up with something better open a new one. Thank you for your efforts.

tr7zw:
Yea I've seen the dev builds and new releases, don't think I will be able to get around to try to make anything that can compete, since I'm tried up with other projects.

PR #67, opened 2022-07-11
Add Imageless FrameBuffers + enable FMA capability for Matrix operations

Vcmp:
- Enable use of Imageless Framebuffers:  (May allow for slight optimisations such as use of a single framebuffer and possibly bypassing the need for explicit transitions in some cases, but I am not entirely sure about the latter)
- Add LWJGL/JOML Performance and Debugging options: (Including enabling FMA which may give a slight performance improvement in some systems)
- Use the newer Load/Store Op None flags via the VK_EXT_load_store_op_none extension: at least on Nvidia it seems to bypass a renderpass read that occurs even if Don't Care is specified, which in this case may not be needed as AFAIK the framebuffer is overwritten each frame anyway, however I haven't found a way of hooking RenderDoc to the Client reliably so I can't confirm this.

Adding Imageless framebuffers should be OK in most cases, as it is very well supported as long as the driver supports Vulkan 1.2: https://vulkan.gpuinfo.org/listfeaturescore12.php

The FMA optimisation however is more of a risk as it will likely cause hard crashes if the CPU is particularly old and does not support the FMA3 instruction set, and I also couldn't find a reliable method of evaluating it during runtime, so I have no issues with removing it if it causes problems of any kind and/or the performance gain doesn't justify it.

I haven't benchmarked this properly yet but there does not seem to be any majorly notable performance regressions thus far.

### DIFF
--- a/src/main/java/net/vulkanmod/mixin/debug/DebugHudM.java
+++ b/src/main/java/net/vulkanmod/mixin/debug/DebugHudM.java
@@ -3,6 +3,7 @@ package net.vulkanmod.mixin.debug;
 import net.minecraft.client.MinecraftClient;
 import net.minecraft.client.gui.hud.DebugHud;
 import net.vulkanmod.vulkan.DeviceInfo;
+import net.vulkanmod.vulkan.Vulkan;
 import net.vulkanmod.vulkan.memory.MemoryManager;
 import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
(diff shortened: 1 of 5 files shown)


SpidFightFR:
Hey, i tested your build, unfortunately, i wasn't able to launch the game.
Here are the logs, if you want:
https://mclo.gs/IxldfmE

Good luck, mate !

LegacyGamerHD:
This will break on pojav

Vcmp:
This should also fix the issues on Linux as some Linux drivers, namely the RADV Drivers, support Imageless Framebuffers but not the VK_EXT_load_store_op_none extension:
https://vulkan.gpuinfo.org/listdevicescoverage.php?platform=linux&extension=VK_EXT_load_store_op_none&option=not

Instead I might think about looking into other things that don't focus on cutting edge driver support, e.g. attempting to fix V-Sync which is currently completely broken in the current version of the mod however that will likely take awhile as it looks like it isn't a trivial issue to fix, (I suspect it is a fence/SwapChain related issue when the Window is in Exclusive Fullscreen but I'm not 100% sure about that)

PR #71, opened 2022-07-15
Fix VSync + Fix Uniform Buffer Memory Leak/allocation related crash

Vcmp:
Added Features:
- Add Basic VSync capability/Fix FrameTime stuttering when in Exclusive FullScreen
     - This was accidentally fixed by messing around with using a separate dedicated transfer queue to Upload all textures/Data instead of the primary Graphics queue which seems to allow for some simultaneous/asynchronous transfer capability
- [Unconfirmed] Improved World Loading, Chunk Loading and Rendering speed
     - Side effect of using a dedicated transfer queue where world loading, chunk loading and rendering speed and efficiency seem to be greatly increased with reduced stutter and CPU usage
     - Unable to confirm this currently so this should be taken with a grain of salt

Fixes:

- Fix infinite Uniform Buffer Allocation memory Leak/Overflow Crash
     - This is a very ugly temp fix done by reverting some changes as a result of a major overhaul in how buffers were allocated
     - this specifically effects the UniformBuffers Class as a workaround for a Uniform Buffer Allocation Bug where the allocation size would be doubled until the GPU would run out of VRAM and cause a Hardware Crash
- Fix graphical artifacts on World Loading screen
     - Also seems to fix a graphical issue where the world loading screen would become bugged and start glitching out shortly before the above crash would occur

Bugs/Issues:
- Seizure Warning: 
     - If VSync is toggled in Exclusive fullscreen, at least on Windows the Screen will show Odd graphical Artifacts such as Distorted/Flashing colours which may cause a seizure risk in some people, unfortunately I am too much of a Vulkan/Mixin noob and haven't found a fix for this currently
     - a temporary work-round is to exit fullscreen and turn ON/OFF VSync (at least on Windows) when in windowed mode 
- Queue Issues
     - This will not work if the device only has one queue as the VSync Workaround requires at least two to function properly, so some devices (e.g. [Google Pixel XL 3 ](https://vulkan.gpuinfo.org/displayreport.php?id=10962#device) and iGPUs such as [Intel HD Graphics 520](https://vulkan.gpuinfo.org/listreports.php?devicename=Intel(R)%20HD%20Graphics%20520%20(Skylake%20GT2))) will not be able to use VSync properly
     - Devices with two or more queues such as [Samsung Galaxy A12](https://vulkan.gpuinfo.org/displayreport.php?id=15772) and [AMD RX Vega 10](https://vulkan.gpuinfo.org/displayreport.php?id=5596#device) should be OK through
- [Unconfirmed] Random Crashes and White Screens (May be a nonissue) 
     - This PR also uses some incredibly stupid performance hacks which are almost completely untested and may either lead to cases of random breakage, glitches and/or do almost absolutely nothing so I don't mind if they or anything else needs to be removed for any reason


Vcmp:
Their seems to be several critical issues including a memory corruption/access violation issue that sometimes occurs when generating and reloading the same chunks several times, so I won't reopen this until it has been cleaned up.

PR #104, opened 2022-09-04
Add additional Info to DebugHUD

Vcmp:
This is a very small Pull Request that makes some minor changes the DebugHUD/GUI which adds the maximum supported Vulkan Version and the GPU Driver version to be displayed.

The Driver version is decoded via with specific bit shifts which coverts the raw Driver version into a human readable driver version number, and the Vulkan version is obtained with vkEnumerateInstanceVersion which can be used before the vkInstance is initialised.

The only issue I am currently aware of is the Driver Version decoder will likely not work with Intel Drivers on Windows and potentially highly Obscure/Exotic GPU Drivers/vendors that use different version formats (As the Driver Version this is only guaranteed to work and display correctly on Nvidia ATM)

### DIFF
--- a/src/main/java/net/vulkanmod/mixin/debug/DebugHudM.java
+++ b/src/main/java/net/vulkanmod/mixin/debug/DebugHudM.java
@@ -59,6 +59,7 @@ public abstract class DebugHudM {
         strings.add("CPU: " + DeviceInfo.cpuInfo);
         strings.add("GPU: " + DeviceInfo.deviceName);
         strings.add("Driver: " + DeviceInfo.driverVersion);
+        strings.add("Vulkan Version: " + DeviceInfo.vkVersion);
         strings.add("");
 
         return strings;
--- a/src/main/java/net/vulkanmod/mixin/debug/DebugHudM.java
+++ b/src/main/java/net/vulkanmod/mixin/debug/DebugHudM.java
@@ -58,9 +58,9 @@ public abstract class DebugHudM {
         strings.add("");
         strings.add("VulkanMod " + getVersion());
         strings.add("CPU: " + DeviceInfo.cpuInfo);
-        strings.add("GPU: " + DeviceInfo.deviceName);
-        strings.add("Driver: " + DeviceInfo.driverVersion);
-        strings.add("Vulkan Version: " + DeviceInfo.vkVersion);
+        strings.add("GPU: " + Vulkan.getDeviceInfo().deviceName);
+        strings.add("Driver: " + Vulkan.getDeviceInfo().driverVersion);
+        strings.add("Vulkan Version: " + Vulkan.getDeviceInfo().vkVersion);
 
         strings.add("");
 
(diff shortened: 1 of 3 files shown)


SpidFightFR (commented):
[It seems to be working.](https://user-images.githubusercontent.com/60798703/188320020-a2ed284c-3563-418e-bfd7-77fc839346a7.png)

Tho i'm not sure that activating Vulkan validation layers is a good idea, because they make the framerate slow and logs are spammed with validations and warnings.

Vcmp:
I just realised I made a slight mistake with the Nvidia version number (Incorrectly displays 515.65.0.1 and not 515.65.10), so I've tried to adjust the Byte shifts slightly, however I am not the best at Byte manipulation so the digits may not be in the right order

xCollateral:
I am almost sure that the version number is referred to vulkan driver version, not vendor specific gpu driver.
https://gpuopen.com/learn/decoding-radeon-vulkan-versions/

SpidFightFR:
> I just realised I made a slight mistake with the Nvidia version number (Should display 515.65.01 and not 515.65.1.0), so I've tried to adjust the Byte shifts slightly, however I am not the best at Byte manipulation so the digits may not be in the right order

Hey ! i see that you disabled validation layers, thank you, it is way smoother now.
Now i think the issue is [not quite fixed yet](https://i.imgur.com/udd9Mb5.png), unfortunately... Keep up the good work pal, and i wish you the best of luck !

Vcmp:
Thanks for the information, this niggle is partially due to how Nvidia handles version numbers on Linux compared to windows as Linux uses XXX.XX.XX (Major - Minor - Patch Numbers) instead of XXX.XX (So the Linux Driver 515.65.01 would look like 515.65 if it was released on Windows) [Can be seen Here: -](https://developer.nvidia.com/vulkan-driver) [and Here](https://old.reddit.com/r/vulkan/comments/fmift4/how_to_decode_driverversion_field_of/fl4mx0t/)

To Simplify Nvidia use a different format for their Driver numbers (10 | 8 | 8 | 6  Bit Shifts) than Vulkan/AMD and most vendors (10 | 10 | 12) so has to be converted differently ([According to Here](https://old.reddit.com/r/vulkan/comments/fmift4/how_to_decode_driverversion_field_of/fl4mx0t/))

I don't have a lot of time currently as I am extremely busy irl atm, so I think I will just take the lazy route for the time being and omit the patch number on the end, so it is more consistent with the Windows system and isn't misleading due to buggy/invalid or incorrect numbering

Vcmp:
I also had the idea of adding other information (e.g. [Pipeline Statistics](https://github.com/SaschaWillems/Vulkan/tree/master/examples/pipelinestatistics) to show Vertex, Active buffers loaded in Host/Device memory e.g.) however that probably won't ever get done very soon if at all, so it will likely materialise in a separate Pull Request if that ever happens

tst

Vcmp:
I haven't gotten around to working on this over the past several days so I will close it for the time being.

Issue #46, opened 2022-06-28
Incompatible and compatible list of mods


Zorklis:
### Testing on VulkanMod 0.4.4 (on 1.21 MC/Fabric.) underway.

Hopefully my testing is useful to someone. Some mods might not function properly but these are the ones that didn't crash on world load. 

### **!!!Comment which mods you tested but are not listed here!!!** 

or not. I'm just a sign.

**Incompatible:**
=========
- autohud-6.6.2+1.20.4-fabric (Doesn't crash in main menu/ Crash during world load)
- astrocraft-1.3.2+1.21
- bedrodium-0.2.3 (requires sodium)
- better-clouds-1.5.7+1.20.4 (crash/ Marked as incompatible.)
- dark-loading-screen-1.6.14 (crashes, also lists other mods as caused by)
- Dashloader
- early-loading-screen-0.5.0
- essential_1-3-1-3_fabric_1-20-4
- exordium-fabric-1.2.1-mc1.20.4 (crash upon world load)
- FastMove-1.20.2-1.0.7 (Uses ukulib: ukulib 1.0.1+1.20.2, which causes crash)
- iris-mc1.20.4-1.6.17 (requires sodium)
- indium-1.0.27+mc1.20.1 (requires sodium)

- ImmediatelyFast-Fabric
- ModernUI-Fabric-1.20.4-3.10.1.5-universal
- Raised
- reeses_sodium_options-1.7.2+mc1.20.4-build.102 (requires sodium)
- particle_core-0.1.2+1.20 (crashes, also lists other mods as caused by)
- PickUpNotifier-v20.4.2-1.20.4-Fabric (Crashes in-game.)
- Ping-fabric-1.21-1.12.0 (Only crashes when used)
- sodium-fabric-mc1.20.2-0.5.3
- sodium-extra-0.5.3+mc1.20.2-build.114  (requires sodium)

- Xaeros_Minimap_24.0.3_Fabric_1.20.4 
- XaerosWorldMap_1.38.1_Fabric_1.20.4
- XaeroPlus-2.6+fabric-1.20.4-WM1.38.1-MM24.0.3

?Tiefix

**Compatible:**
=========
+3dplaceablefoodUPDATE-1.1.8 1.20.4 fabric
+abridged-fabric-1.20.x-1.1.1
+abseil-fabric-0.96.4+1.20.4-1.0.1
+accelerated-decay-fabric-84.0.0
+accurate-block-placement-1.2.1
[...]

PR #107, opened 2022-09-18
hard fail on incompatible mods

ArnaudLier:
this gives a better explanation to end users when a mod from #46 is loaded than just a crash
(also removed the default `suggest` dependency)

### DIFF
--- a/src/main/resources/fabric.mod.json
+++ b/src/main/resources/fabric.mod.json
@@ -31,7 +31,15 @@
     "minecraft": "1.18.x",
     "java": ">=17"
   },
-  "suggests": {
-    "another-mod": "*"
+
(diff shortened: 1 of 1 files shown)


Fox2Code:
I usually don't like hard fail on incompatible mods unless strictly necessary, because this prevent the other mods from ever being compatible with VulkanMod.

At this time Fabric don't expose any APIs to allow a mod inside "breaks" to start, so it may ruin indefinitely the change of any of theses mods of working with VulkanMod ever.

ArnaudLier:
What about only hardfailing on mods that will never be compatible with VulkanMod like Sodium?

Fox2Code:
Yeah I agree for that.

PR #109, opened 2022-09-24
Use Precompiled Shaders

Vcmp:
This attempts to reduce loading times by precompiling the shaders as int arrays which are then embedded into separate classes and loaded directly 

The main issue with this change is that the GLSLC utility needs to be on the Path as an environment variable so that the 'genFrag' or 'GenVert' gradle tasks used to update the Frag and Vert classes can call it on the command line, [glslc is included in the Vulkan SDK](https://vulkan.lunarg.com/sdk/home) (which is normally not needed as the SDK is usually C/C++ specific and it's functionality is bundled with LWJGL)

The other problem is that the aforementioned gradle tasks I added are extremely crude and not well written so they will only run if the source shader .vsh and .fsh files are updated or changed, they are also incredibly ugly and cause alot of clutter and bloat in the gradle build file. (This does work well when wanting to make changes/optimisations to the shaders and quickly applying them to check changes in game however)

There is likely a better way this could have been done (e.g. loading .spv files instead of int arrays through invoking a java submodule instead of a command line executable), however I wanted the shader loading to be as fast as possible as I wasn't sure of the performance penalty of file management/opening and the risk of missing files causing issues.


Vcmp:
The latter two 'Test' commits were just testing changes made to the shaders were being reflected and compiled correctly in the Shader Classes, so they can easily be removed if they cause issues/performance regressions.

Also forgot to mention earlier that the glslc compiler used for this uses additional flags including the '-O' optimisation flag which afaik was not applied originally, so there may be very slight performance improvements, however this is completely unconfirmed.

SpidFightFR:
Hey, using precompiled shaders seem to give better performances, tho, i have issues with [the sky](https://imgur.com/a/7eejagD), it doesn't refresh properly, the sun and moon both move like they should, but the sky color alongside stars while at night don't refresh, staying in place when we look around, also making weird trails at some part.

I don't know what's causing this...

SpidFightFR:
also, is it possible to compile the shaders ourselves, for our hardware, similar to [fossilize-replay](https://github.com/ValveSoftware/Fossilize) ? Or is it useless, or impossible ? (I'm a noob, in that domain, so sorry if that sounds dumb!)

SpidFightFR:
Also, i tried re-compiling your modified fork, first time went well, tho since second time, [the game crashes on startup](https://mclo.gs/j6HbNCz), i tried with quilt, fabric, and creating a new instance, it doesn't seem to work.

xCollateral:
I've been thinking about this too. This is a lot of work, good job. Though I was thinking about doing something easier i.e. setting a static public flag for compiling or not (like the one used for enabling validation layers), so that when developing it would be easier to chose to compile shaders or not.

Fox2Code:
Well, I think making the game compile shaders only once, then cache the compiled result in a folder could be a better alternative, this will allow both the mod to load faster and to be modular.

I like the code generation part, but I think it should be generated on compile to avoid git conflict when editing shaders.
(If this is merged I'll handle that part myself later)

Also instead of using `/temDir` using `/build/temDir` is better as `/build` is auto managed by gradle.

xCollateral:
@Vcmp Any news on this? If you want you could join discord server and talk there.

Vcmp:
(Apologies for the lack of activity as I was extremely busy IRL for several weeks)

the code generation was a technique I came up with to allow GLSL to be converted to int arrays containign the compield SPIR-V code, However a main issue with this fork is that the shaderc embedded library bundled with LWJGL lacks the ability to output compiled SPIR-V as int arrays, as only the glslc utility has that ability.

This also makes building the mod clunky as glslc.exe must be on the commandline. The gradle task to assemble the shader class files is also buggy and if glslc.exe is not found/the task fails for any reason, the generated class becomes empty and leads to the game instantly crashing.

The Gradle tasks that act as Shader Class Generators also lacks modularity and can't easily add/remove or HotSwap/HotReload shaders (e.g. similar to the Optifine System based on the [GlslShaders mod from years ago](https://www.minecraftforum.net/forums/mapping-and-modding-java-edition/minecraft-mods/1286604-shaders-mod-updated-by-karyonix)). 

A possible workaround is putting each compiled shader into a separate class to allow modularity, which will involve dynamic class loading (although apparently you can use ClassLoaders to load precompiled classes, however that might impact load time)

An easier workaround is to compile the shaders as raw .spv files and loading them as Little-Endian ByteBuffers, which was initially avoided at the time as initially I wanted to keep the system of using compiled int arrays. I will try to find other workarounds, but failing that file loading can be used as a last resort.

xCollateral:
I think you should divide PR in 2 PR, one for shader optimizations, one for shader pre-compilation

Vcmp:
The fork has been changed so that it uses separate classes for each shader, the class loading is handled with _Class.forName()_ which seems to use reflection to allow for the shader array to be accessed directly. 

So in theory if a specific ShaderName is provided and a class with a matching name is found, it should in theory be able to load the shader into the specific pipeline's shaderModule it is being provided to before calling _vkCreateShaderModule()_, as long as the class exists beforehand and the names are identical.

The problem with huge files being added to git is still an issue, however I wanted to leave them in for the moment for testing purposes until I find a better method of handling the shader compilation without awkward to use commandline utilities.

Vcmp:
(This also does not make any changes to the shaders themselves as they are still identical, just the way they are loaded is changed to try to make them more modular)

xCollateral:
I think it would be better to just use shaderc, setting a global flag for compiling or not shaders

Vcmp:
That makes alot of sense as I had convinced myself that [GLSLC](https://github.com/google/shaderc/tree/main/glslc) compiled faster shaders than the specific shaderc version bundled with LWJGL, however the current setup is far too awkward and prone to breakage.

 So I will try to replace it with a much less complicated method using shaderc instead to handle the compilation and writing them to disk as compiled SPIR-V files and loading them when needed.

xCollateral:
@Vcmp If you need help or anything come chat in our discord server, here's invite link: https://discord.gg/FVXg7AYR2Q

xCollateral:
yes it has and i have tested it (there's a commented line on `ShaderSPIRVUtils` class) and i didn't notice any difference in performance, maybe these shaders are too simple to get a significant benefit.

PR #120, opened 2022-10-09
Precompiled shaders & shader optimizations

Vcmp:
Similar to the previous PR, but uses a much less complicated system where Precompiled shaders are loaded through saved files instead of through baked class structures and avoids needing to install complicated to use tools to use properly.

~To compile the shaders normally, the _RECOMPILE_SHADERS_ boolean in the Vulkan.java file needs to be set to true when building the mod for the first time to avoid the game crashing on startup, afterward it should be OK to set _RECOMPILE_SHADERS_ to false for every subsequent run unless the shaders are modified in any way.~

Edit: this isn't necessary anymore as the bool will be overridden if the folder isn't there so the bool only needs to be changed if the shaders are updated/modified in any way.

One of the main issues compared to the other PR is that the shaders are not updated automatically if they are changed, so the bool will need to be toggled every single time any changes are made so they are applied properly.

The other issue is that I'm not 100% convinced performance is exactly the same as the other PR (Loading/Framerate may be slightly different)


SpidFightFR:
Hey i tried to compile your PR, the game [crashes on boot.](https://gist.github.com/SpidFightFR/2c0aed17c921f34191ccc35d58f3c0a7) Crash report attached.

Vcmp:
Sorry about that, I forgot I had -Dorg.lwjgl.util.NoChecks=true enabled as a VM Option/Argument. The issue should be fixed now.

(the crash was caused by the missing instance field in the VmaAllocatorCreateInfo object where if checks were enabled LWJGL  would cause a crash for debugging purposes if the code had any mistakes to avoid possible bugs)

SpidFightFR:
Yo, i compiled your new commit, and it works ! Great job !
Also the previous issue with the sky not refreshing is solved !

xCollateral:
I just pushed a commit that fixes shader compile time in lwjgl 3.3.1 by initializing once the shaderc compiler instead of every time( probably it was a lwjgl 3.2.3 bug and not 3.3.1, i suppose compiler wasn't really destroyed).

xCollateral:
Looks good. Please revert struct declaration in `Pipeline` class to older style to mantain same format in the whole repo.

Vcmp:
The "Improve Compilation Speed" commit was reverted as I didn't realise realise that the _shaderc_compiler_release_ fix was already added, the only other major differences being cleaning up my sloppy coding style and that the Shaderc/glslc Vulkan and SPIR-V version used when compiling is set to the maximum version the GPU Driver supports instead of always vulkan1.1 in case it helps with performance slightly.

Vcmp:
Unfortunately I can't seem to reproduce that crash, it either is due to using outdated shaders in the "run/compiled" folder or it is failing to load the shader files correctly for some reason.

What might work is deleting the "compiled" folder in the run directory in the project/game folder or failing that to change the "RECOMPILE_SHADERS" bool to true in vulkan.java at line 57 to regenerate the shaders (as it is likely loading the older shaders which will not load properly)

Vcmp:
I tried generating and using a mod artifact and can confirm it works at least on Windows 10 Ver 21H1 build 19044.1766 with MultiMC, I have added a copy of the version I used in case it is useful.
[VulkanMod_1.18.2-0.1.8.zip](https://github.com/xCollateral/VulkanMod/files/9759771/VulkanMod_1.18.2-0.1.8.zip)

Vcmp:
Just in case it helps I have attached a copy of the jar artefact I used earlier which is confirmed to work at least with MultiMC on windows [here](https://github.com/xCollateral/VulkanMod/files/9759771/VulkanMod_1.18.2-0.1.8.zip). This uses LWJGL 3.2.2 shaphot+4 instead of 3.3.1 but is mostly identical outside of that.

If all else fails I will try to roll back to the last commit/version that works to avoid any more issues, which isn't as bad as it sounds as most of the recent changes are either overkill or not that important (mostly quality of life changes to try to help make the mod easier to code) and don't seem to give much performance gains anyway. 

I think I tried to push the optimisations too hard so I will scale them back if it becomes a problem.

SpidFightFR:
@Vcmp Hey, i tested your latest build ! No signs of Errors whatsoever, tho the compiled shaders seem to be "oneshot shaders": you can only use them once, then you have to delete them once you close the game. Otherwise the game crashes on launch.

Besides that, good job, as always! ;)

Vcmp:
I might have figured out the cause for the crash issue you mentioned earlier, (the ShaderLoader class I added doesn't always properly free the bytebuffers used to setup the shaders), (e.g. I think on windows they are freed automatically while on Linux they are not.) which explains why it was crashing on Linux but not Windows.

You can try using the latest commit just to see if it actually fixes the crash, but I'm not sure if this build/PR makes much sense anymore as it only cuts off a few seconds on the load time (As the long title screen load time bug this PR was designed to workaround was fixed a while ago) and at least on my machine the performance improvement is very minor (about +10%)

However this may be higher on a different GPU, especially if the CPU is faster than my current rig which is ancient (i7 4790K which is over 8 years old) so if the performance gain is different/higher on your./other people's end I may keep this PR open if it actually makes a meaningful difference.

Vcmp:
I just noticed that the last two commits cause validation errors/incompatible format warnings due to changing the formats of the Vertex attributes used by the pipelines which is a problem due to the formats being hardcoded in the game which the mod cannot override very easily without involving hacky/messy code to do so. 

This doesn't cause the game to crash but it will likely need to be dealt with as it may cause issues with other Graphics Card drivers so I will split this into two separate PRs to avoid this one becoming any more of a mess that it is currently.

SpidFightFR:
Hey, the last build solves the issue with keeping the shaders on linux !

Vcmp:
Looks like it actually was a memory issue specific to Linux, so I'm glad that fix seemed to work, does this fork actually improve performance in-game or is it about the same? 

There were some other shader tweaks I tried to add but they ran into validation errors that would involve very invasive changes to the mod to fix so I had to revert them.

SpidFightFR:
Loading times are equivalent (maybe one or 2 seconds more or less but it’s not really noticeable).

andreasdc:
What will this PR bring btw?

Vcmp:
(tldr: PR consisted of two things: reduce long load times, and shader optimisations to improve framerate slightly, the former is mostly redundant as the slow loading bug was fixed upsteam recently, and the latter I'm not 100% sure actually makes much of a difference if any)

All this PR was originally intended to do was to fix an annoying issue with LWJGL 3.3.1 where the shaders would take forever to compile which would cause the game to take about 20-30 secs to reach the title screen (an why the mdo was stuck with LWJGL 3.2 for a while). This was done through precompiling the shaders instead and loading them as files to skip the slowness during loading and speed up the loading process.

However in practise this PR doesn't make much sense anymore as the cause for the slow load times was found and fixed several days ago, and while it does still help a bit it only saves a few seconds of the load time so it isn't really needed anymore.

The other changes added was attempted shader optimisations to in an attempt to improve the framerate/performance but whether it actually makes a tangible difference is debatable as I don't have a good method of benchmarking currently to verify this.

andreasdc:
Can you improve performance leak caused by nametags/entities rendering?

SpidFightFR:
i don't think that's related to shaders... try opening an issue about this if it's not already the case.

Vcmp:
I didn't know their was a memory/performance leak with nametags/entities. As mentioned above I highly doubt it is a graphics related issue and more due to the game itself adding massive CPU overhead when handling the entities which is completely separate with anything to do with Vulkan. 

I can try to look into it later when I have some time but if it isn't a graphics/GPU performance related issue that may likely be better handled by other mods as IIRC this mod is mostly focused with overhauling the rendering engine and rendering itself and minimising changes to other aspects of the game itself as much as possible to help it be as compatible as possible with other mods.

andreasdc:
https://github.com/xCollateral/VulkanMod/issues/127

Vcmp:
Sorry my apologies, that actually does look like an issue that is worth looking into as the Rendering calls in Vanilla are some of the worst CPU hogs in the base vanilla game. I will try too look into it later will open a new PR if I can figure out a way of cutting down the CPU overhead.

andreasdc:
> Sorry my apologies, that actually does look like an issue that is worth looking into as the Rendering calls in Vanilla are some of the worst CPU hogs in the base vanilla game. I will try too look into it later, and will open a new PR if I can figure out a way of cutting down the CPU overhead.

I can't wait to test it :) but the latest version breaks my cutom gamma, so i can't use it :(

andreasdc:
I think I have the version from this commit https://github.com/xCollateral/VulkanMod/commit/b7e02eec4c19df4c535f83c0fe5f5029c3cc1143

Vcmp:
Also I'm not 100% what is meant by gamma (I think its related to black levels) but that might also be a Shader/ColourSpace issue which probably isn't very hard to fix. I am aware of a screenshot bug where the RGB/BGR colour channels are swapped, however that is a separate issue most likely.

andreasdc:
The gamma is brightness, I put custom value in the options.txt file, but Collateral added settings gui which brings back my gamma to 1 instead of what I put in the file.

Vcmp:
I'm 90% sure that most of the in game graphics parameters (fog start-end, lineWidth, endPortalLayers e.g) are controlled from mixins that hook into the game and gets the values manually. Its possible that some of the values, including the brightness one  is hardcoded to a fixed value/or isn't being updated properly and is missing a dedicated update hook for it.

Vcmp:
I think I will close this PR as while the Shader precompiling/Preloading made sense when the long shader compile times bug/issue was a factor with LWJGL 3.3.1 specifically, as that has been fixed upstream, the loading time advantage added by this PR is only roughly a few seconds compared to the main/upstream branch and doesn't have any tangible performance improvements afaik.

Additionally, as the shader/other optimisations added don't seem to give any major or game changing improvement, (may not even be 10%) I think overhauls like that are best spent on other areas that will likely give much greater gains (e.g. looking into the CPU bottleneck with rendering ATM instead of improving GPU performance/efficiency as the game is mostly CPU than GPU bound currently with any card that's roughly faster than a RX 580/GTX 1060 ish)

xCollateral:
I undestand what you say. Deeper changes to the rendering engine are needed to really make a difference in performance.
Anyway i wanted to thank you for all the work you have made.

SpidFightFR:
Thank you, for that pr.

Vcmp:
Apologies if that sounded rather solemn, I'm not nessacery throwing out what was done in this PR, I get side-tracked very easily and as a result I started to get the impression this PR had turned into a mad experiment that was going down a rabbit hole it didn't need to go down and wasn't really going anywhere very quickly.

In other words I decided that it was much better to take it it in a more productive direction potentially recycle parts of this PR and look into other areas instead that actually made sense (As many of the issues seem to come from oddities in the vanilla code and not this mod)

Vcmp:
In other words I decided that it was much better to take it it in a more productive direction (potentially recycling parts of this PR as well) and look into other areas instead that actually made sense as this mod is not the problem. (only uses about 5-10% of the CPU with Vanilla using the rest).

andreasdc:
It sounds like you lost some motivation, if you see the sense and you have the idea, you should make it, I keep my fingers crossed for you! ❤️

PR #168, opened 2022-12-31
Try to optimise Chunk rendering

ghost:
This PR is an experimental attempt to improve the performance of chunk rendering through addressing some issues made in how chunk rendering is handled. (mostly consists of a mix of workarounds and simplifying chunk rendering (single shader instead of 5 is used to render chunk VBOs, which reduces drawcalls by roughly 80%))

As a result it doesn't fix the poor performance with Mobs and Block/Tile entities, so it may be useful to disable mob spawning to reduce performance bottlenecks. AFAIK It also doesn't perform as well as the earlier 1.18.2 version this PR was based from due to rendering in 1.19.2 being slower compared to 1.18.2.

There are way too many features and changes made to summarise or explain properly, so I don't mind splitting this PR and/or removing unnecessary changes if needed.

As this is highly experimental there is a chance of random breakage/crashes occurring randomly for no reason. AFAIK there are no major crashes that exist currently but there are some minor graphical glitches. (The current glitches include Water/Rain/Translucent particles being buggy, Block breaking texture and water rendering being completely broken, and temporary Optifine-Style Farlands glitches if teleporting too far e.g.)


ghost:
This PR also has a problem where the performance tanks in Mangroves/Jungle Biomes, however that seems to be a long standing Vanilla bug as jungle lag is nothing particularly new.

It also causes a weird issue where the CPU usage with vkWaitForFences gets extremely high (over 60%) in mangroves particularly. But if you go into a Biome with no leaves/vines performance seems to go back to more normal levels.

So whatever is causing the Leaves/Vines lag and effecting Vanilla/OpenGL is also effecting Vulkan as well, and therefore might be something this mod can control/fix to some extent. 

However I won't worry about fixing this too much, as AFAIK it is mostly a vanilla issue and is likely unrelated to this PR anyway.

ghost:
Another issue is that this PR uses at worst about 40-60% more GPU memory, (in a single biome mangrove world, the main 1.19.2 branch used about 1GB of GPU memory while this PR uses about 1.6GB). 

This is likely due to the buffer management system used in this PR, (which is completely different to that used in the main branch), not being as effective with freeing unused chunks properly and wasting space on the GPU. 

This is related to this PR unlike above, so I will try to find possible fixes to cut the usage down to more manageable levels (to allow older/lower-end GPUs with smaller memory capacities to be viable)

xCollateral:
Wow, you've done quite a lot of work, I appreciate your effort. I checked out this PR and there are many aspects that I wanted to implement myself. VBO areas is one of the most promising.
I tried running it but it crashed on startup due to a timeout error on _vkAcquireNextImageKHR_ call. I noticed the timeout time was too short so i reverted it back and got it running.
It crashes shortly after loading a world with _VK_ERROR_OUT_OF_DEVICE_MEMORY_, so there are many instability issues.

I suggest you not to make unnecessary changes, try to maintain code style and structure as much as possible, even splitting this PR could help.
I think coordination is needed for such a deep change in rendering to make sure we go in the same direction.

ghost:
That makes sense as this PR came from a stupid idea I had to reduce lag at high render distances, the PR also has a lot of issues including ugly hacks and unnecessary code I'm not 100% happy with myself and its performance not being as high as I wanted initially. The original version of this PR was on 1.18.2, which I then later badly ported to 1.19.2 with inferior performance to the former, which is likely where some of the ugly code likely comes from.

What could work is splitting the PR into the VBO/Shader changes and the Virtual Buffer Changes separately and removing other changes unnecessary for the mod. It might be difficult to split it further due to some of the changes being dependent on one another. 
(Bindless rendering needs the Virtual/UberBuffer System, which itself needs the single shader setup for VBOs to avoid splitting Meshes based on RenderTypes so they can be packed into a single buffer).

The Swapchain/VK_ERROR_OUT_OF_DEVICE_MEMORY issues you ran into are likely partially Driver/OS/Hardware specific as this runs fine and is very stable on my GTX 1080 Ti on Windows 10, although I use an outdated Vulkan Beta Driver and my PC Rig is ancient, so that may be a factor. (I also don't have an AMD card to test with).

Outside of that the other likely cause is the reserved memory resizing code not working particularly well, I set the initial size far too low, so I wasn't surprised it caused crashes when resizing the Buffer due to running out of space. (to avoid overwriting VBO meshes). I'm 99% certain that is the cause as that is the only place that specific VkResult is thrown outside of the Main Branch Code.

A possible crude workaround is simply setting the reserved size to something like 2GB and never resizing it again but I didn't want to do that at the time as it is wasteful and not all GPUs have 2GB available.

ghost:
I've just added some temp fixes to get the PR in a more workable state ATM. (Main changes are disabling getting multiple images from Swapchain and setting the Contiguous buffers to 2GB as a temporary measure). 

I can't set it much lower due to extreme edge cases of complex Geometry (using a single Biome Jungle/Mangrove World at Full 32 Render Distance.) Using lower Geometry Complexity (e.g. SuperFlat World) may help if the VirtualAllocator is still causing issues.

I also don't mind collaborating to get aspects of this in a more workable state and to drop or remove anything unnecessary, as their is alot of ugly code and additions that are not needed.

The Virtual/UberBuffer system used in this PR also has alot of Issues with memory, especially Resizing, so i'm not 100% convinced its fully workable for old/low-End GPUs with smaller memory capacities.

This PR was also supposed to have Draw Call Combining (Merge all VBOs into a single Mesh and render in one Draw Call) however I had problems enumerating and lining up the correct Vertex/Index Counts and Offsets for each VBO Mesh and handling the fragmentation gaps properly so it was not added.

ghost:
Realised I made a silly mistake with multiDrawIndirect where I wasn't using it properly (Was uploading the DrawCommandsArray every frame when it is only needed to upload them if the VBOs change,) which was causing an approx. 20% performance hit If multiDrawIndirect was enabled. 

After fixing this, using multiDrawIndirect seems to now give 10-15% performance gain if enabled, when it was 10-20% slower before. (At least on my setup, prob varies based on OS, GPU Driver, CPU, RAM e.g.)

Pevernow:
This PR is great.
I'm also trying to add some useful performance optimization for this Mod
For example, FSR2 and DLSS support
Maybe I will try to add hardware ray tracing support later (experimental)
I'm looking forward to seeing this PR continue to develop

PR #177, opened 2023-01-15
Fix screenshot colours being packed as BGRA instead of RGBA

ghost:
This is a quick and dirty PR to attempt to fix a long standing bug with the mod where the colour channels were being saved in the incorrect layout (BGR instead of RGB) leading to messed up colours when they were saved.

This does not fix Vanilla Behaviour of the Alpha channel being saved in the screenshots however, so image file sizes will still be about 50% higher than they normally need to be (Due to the alpha channel adding an extra byte so 4 bytes per pixel instead of 3 with RGBA and RGB respectively).


ghost:
Realised this might work alot better with a compute shader to strip out the alpha channel, so will close for now to test setting it up to see if it has any advantages over blitting.

ghost:
The Blit has been changed with a computer shader pipeline instead which should hopefully work a little better. 
This also changes how the alpha is handled by backfilling each RGBA pixel with 255 max Alpha to avoid weird Alpha Blending issues with some image viewers.

PR #186, opened 2023-02-16
Fix messed up  screenshot Colours (Saving as BGR instead of RGB)

thr3343:
_(Had to close my old account due to some privacy issues hence why this is reuploaded)_

This PR attempts to fix an old bug in the mod where the screenshot colours are not converted to the correct RGB pixel format prior to saving the Framebuffer/Swapchain image as a .png file.

This is done through a separate Compute Shader to Swizzle the Pixel colour channels when saving/handling the screenshot, and due to how the compute step is handled should have no effects on gameplay or performance in any tangible way.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,7 +8,7 @@ org.gradle.jvmargs=-Xmx1G
 	loader_version=0.14.8
 
 # Mod Properties
-	mod_version = 0.2.0+1
+	mod_version = 0.2.2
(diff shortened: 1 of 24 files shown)


thr3343:
Will be closing this as the BGRA->RGBA fix was added upstream which works way better and is far less complicated than this PR anyway. 

The PR was mostly an excuse to learn Compute Shaders Anyway so don't mind closing this.

PR #187, opened 2023-02-16
Chunk Rendering/VBO Optimisations

thr3343:
_(Had to close my old account due to some privacy issues, hence why this is reuploaded)_

Attempts to address limitations of the current VBO system used to render chunks through exploiting several FPS/performance boosting methods and techniques, among the most major of these including using a combined renderlayer setup in tandem with a janky/buggy UberBuffer/VirtualBuffer system. 

The former reduces drawcalls by roughly 80% in-game through combining all geometry from the separate renderlayers into one, and the latter allows for faster VBO uploads due to avoiding the need to recreate/delete buffers every frame and additionally provides the ability to use bindless rendering. 
(There are many other minor technical improvements and changes, however there is far too many to list here.)

### Currently known Bugs AFAIK
- **Very rare Vertex Explosions and Graphical Glitches due to offset errors,** (Unconfirmed: May be fixed/not exist anymore)
- **Crashes on AMD GPUs when VBO capacity is exhausted** (Unconfirmed: Don't have an AMD GPU to test)
- **Mipmaps are completely broken,** 
- **High GPU memory usage, (Always uses 2GB)**
  - Is also fixed in capacity and can't resize 
- **Chunks are forcibly reloaded if it exceeds the fixed 2GB limit** (More a technical Limitation with Vulkan than a traditional bug)
- **Mangroves/Jungle Lag,** (Strongly related to an ancient vanilla bug with leaves/Vines)
  - vkWaitForFences() CPU gets very high when rendering mangroves/jungles and causes lag. 
  - (Occurs in upstream as well, but is much worse here)


ezamelczyk:
I haven't been able to even load a world with your changes.

Stacktrace:
```
[11:41:53] [Render thread/ERROR]: Unreported exception thrown!
java.lang.IllegalAccessError: class net.vulkanmod.render.chunk.WorldRenderer tried to access protected field net.minecraft.class_1159.field_21655 (net.vulkanmod.render.chunk.WorldRenderer and net.minecraft.class_1159 are in unnamed module of loader net.fabricmc.loader.impl.launch.knot.KnotClassLoader @77e4c80f)
	at net.vulkanmod.render.chunk.WorldRenderer.renderChunkLayer(WorldRenderer.java:575) ~[VulkanMod_1.19.2-0.2.0+1.jar:?]
	at net.minecraft.class_761.method_22710(class_761.java:4049) ~[client-intermediary.jar:?]
	at net.minecraft.class_757.method_3188(class_757.java:1039) ~[client-intermediary.jar:?]
	at net.minecraft.class_757.method_3192(class_757.java:816) ~[client-intermediary.jar:?]
	at net.minecraft.class_310.method_1523(class_310.java:1177) ~[client-intermediary.jar:?]
	at net.minecraft.class_310.method_1514(class_310.java:768) [client-intermediary.jar:?]
	at net.minecraft.client.main.Main.method_44604(Main.java:244) [client-intermediary.jar:?]
	at net.minecraft.client.main.Main.main(Main.java:51) [client-intermediary.jar:?]
	at net.fabricmc.loader.impl.game.minecraft.MinecraftGameProvider.launch(MinecraftGameProvider.java:461) [fabric-loader-0.14.14.jar:?]
	at net.fabricmc.loader.impl.launch.knot.Knot.launch(Knot.java:74) [fabric-loader-0.14.14.jar:?]
	at net.fabricmc.loader.impl.launch.knot.KnotClient.main(KnotClient.java:23) [fabric-loader-0.14.14.jar:?]
	at org.prismlauncher.launcher.impl.StandardLauncher.launch(StandardLauncher.java:88) [NewLaunch.jar:?]
	at org.prismlauncher.EntryPoint.listen(EntryPoint.java:126) [NewLaunch.jar:?]
	at org.prismlauncher.EntryPoint.main(EntryPoint.java:71) [NewLaunch.jar:?]
```

Alongside there was a bunch of error logs like this:
```
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkAttachmentLoadOp value 1000400000 is not supported.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkAttachmentLoadOp value 1000400000 is not supported.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkAttachmentStoreOp value 1000301000 is not supported.
```

thr3343:
Sorry about that, this has a lot of old, experimental and ugly code that is very different from the upstream VulkanMod code. It was also used on a GTX 1080Ti as I don't have a AMD card to test with, so I'm not surprised their are problems running it.

The above crash is due to some Access Wideners which don't seem to have been applied properly, IIRC the _remapJar_ Gradle task needs to be run first to apply them properly.

The second error are due to some [Vulkan 1.3 features](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_load_store_op_none.html) added for testing which are not supported by that specific GPU. It is technically more efficient but it doesn't help performance in anyway so I will remove it. (Is well supported on Windows but mostly forgot about other platforms)

I have also removed some of the Access Wideners just in case to helps address the issue. 

If there are still issues I can try to provide a mod artefact that can be copied into the mods folder and used as a mod normally would, however that is a last resort as I don't want to distribute unofficial versions of the mod as I'm not sure xCollateral is OK with that.

thr3343:
This contains alot of very ugly and outdated code from an older version of this mod, and also has massive technical limitations and problems which isn't really particularly workable or fully compatible with the mod anymore.

So I decided to clean-up most of the ugly code and move some of the more stable optimisations to a different PR, which isn't as fast or performant as this one, but in exchange has almost no rendering glitches and therefore should be drastically more playable.

The rest of this PR will likely get mothballed and converted into a Prototype/Raytracing testbed, which will not turn into anything significant unless it get completed and/or becomes playable which is unlikely.

PR #192, opened 2023-02-22
Add runtime for mac arm platform

ezamelczyk:
Pretty simple change and it adds compatibility with m1/m2 macs. Built and tested on M1 Macbook Pro. Works pretty good with occasional hiccups. Without that change the game would crash on startup.


xCollateral:
I am aware those libraries are needed to run the mod on MacOS, there are indeed MacOS builds on Modrinth. I prefer to keep them separete due to jar size which is already big enough.

PR #196, opened 2023-03-02
Little Update to README.md

SpidFightFR:
Adding more information in the "known issues" tab.
Removed "Linux may not work", as it do work.
Added a link to the archwiki explaining how to get vulkan to work on linux (it's pretty well explained).

Didn't change the "macOS may not work" as i don't own a mac, i can't really tell if that's still the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,12 @@ Demostration Video:
 
 ## Known Issues
 
-- Linux and macOS may not work.
+- Make sure your hardware supports Vulkan, using for instance [that tool for windows](https://github.com/skeeto/vulkan-test), or using the `vkcube` command on linux to make sure vulkan works.
+- MacOS may not work.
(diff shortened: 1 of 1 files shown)


thr3343:
I'm likely wrong about this as I don't run Linux or have an AMD GPU, but the Linux disclaimer may be referring to the RADV AMD Drivers, which apparently seem to have the most issues with this mod for some reason (Including Swapchain-related bugs apparently).

Vixea:
https://i.imgur.com/0wFDoKm.jpg
@thr3343 tested on RADV works fine now! :D

ghost:
AMDGPU users still having some issues with it, like everytime the mod update something will break and cause driver crashes. We're not sure about it's Mesa thing, but Collateral tried his best to fix it.

Vixea:
Have they tried forcing vsync?

PR #200, opened 2023-03-12
Minor VBO + RenderPass Optimizations (Stable Version)

thr3343:
This is a cleaned up version of the [VBO Optimisations](https://github.com/xCollateral/VulkanMod/pull/187) PR, with most of the ugly code/optimisations heavily cut down. This hurts performance alot, but heavily reduces glitches and improves stability as a result.

In Short, this attempts to reduce CPU rendering Overhead by removing the need to submit PushConstants for every VBO per frame, which can easy reach over a million or so due to PushConstants being uploaded every (FPS*VBOs) per second.

Other addons include minor GPU optimisations and a refactored VBO system, However the latter isn't 100% nessacery and is mostly included to help fix my sloppy codestyle and to improve VBO performance, so I am perfectly happy to remove these and/or make any other changes as needed.

Despite the above, this is still experimental and is likely not Bug/Glitch free and may include graphical glitches ( e.g. a long-range World-Origin teleport-related bug) but it should be mostly playable to my current knowledge.


src/main/java/net/vulkanmod/mixin/chunk/LevelRendererMixin.java
thr3343, line 1: Remove unnecessary clear

src/main/java/net/vulkanmod/mixin/render/GameRendererMixin.java
thr3343, line 1: Remove unnecessary clear

src/main/java/net/vulkanmod/vulkan/Drawer.java
```diff
@@ -181,7 +181,7 @@ public void initiateRenderPass() {
             renderPassInfo.renderArea(renderArea);
 
             VkClearValue.Buffer clearValues = VkClearValue.callocStack(2, stack);
-            clearValues.get(0).color().float32(stack.floats(0.0f, 0.0f, 0.0f, 1.0f));
```
thr3343, line 183: Color can be provided to renderpass directly instead of just clear attachments

src/main/java/net/vulkanmod/vulkan/Vulkan.java
thr3343, line 1: Can use malloc as the color formats are not read if only depth format is specified. (i.e Depth only clear)

src/main/java/net/vulkanmod/vulkan/Drawer.java
thr3343: Apparently your not supposed to use too many clear colors (15-30) at least on Nvidia, which disabled color compression apparently and wastes bandwidth, but in practise saw virtually no difference to Perf what so ever even when forcibly disabling all clears.

thr3343:
This wasn't part of the original PR but decided to also make some RenderPass adjustments to reduce RenderPass/Attachment clears, which apparently can waste bandwidth and disable clear colour compression if you use too many different/unique clear colours.

Colour attachment clears are also disabled, with Depth only clears being used instead.

This also follows AMD recommendations which state that you should always use [Fast Clears](https://gpuopen.com/performance/#clears) which are over 100x faster apparently.

This likely will make little to no difference on most systems but may give a small improvement to Super low-end/Mobile devices/iGPUs that are highly bandwidth constrained.

thr3343:
The only other changes made is that the depth buffer uses Depth only formats instead of Depth+Stencil ones as the Stencil test is literally never used at any point, as a result the depth formats were adjusts in case it gives a small speedup.

The Format will also be adjusted based on which GPU finds the fastest and/or performs the best with, (e.g. Nvidia prefers D24 Bit while [AMD prefers D32-bit](https://gpuopen.com/learn/dcc-overview/))

thr3343:
~Made a mistake with the "Enable OIT + Reduce vkBuffer Uploads" Commit which is misleading/incorrect, it is not actually Order-Independent Transparency that was changed/fixed in that commit and was instead an odd Vertex-Sorting issue where the Translucent VBOs get messed up if there combined with the Opaque/Non-Alpha textures.~ 

~Was dumb and got the two mixed up, hence the incorrect label/commit message, in addition that specific commit did the opposite and broke Transparency ordering and therefore was not actually fixed. (Will try to fix this later that hopefully will not cause any tangible performance regressions)~

Edit: Ignore the above, the "Enable OIT + Reduce vkBuffer Upload" is still wrong as it isn't Order-Independent Transparency, However the Translucent texture/vertex sorting is partially a very subtle bug in VulkanMod that wasn't discovered for a long period of time due to being extremely difficult to notice under normal circumstances: (Until I messed around with this earlier)

Xemorr:
Can we get this accepted into the project, it sounds good to me.

thr3343:
IMHO, this is a highly experimental PR, and shouldn't get merged in its current state as there are still some major graphical glitches, namely Translucent Blocks i.e. Stained glass and Mipmaps, which is completely broken in this PR 

(ATM I'm experimenting with a different method of rendering translucent blocks as they are much slower to render than the other layers, which hasn't been tested yet to see if it actually works or not)

In addition xCollateral is extremely busy IRL including working on other parts of the mod, and likely won't have time to look at this until later in the year, which I don't mind as this PR isn't 100% finished yet anyway.

thr3343:
Will likely close this now as this is old tbh, and some of the optimisations included in this have been added to 0.3.0 and 0.3.1 after working with/discussing with xCollateral about some of the stuff in the PR.

Some of the other features may appear in a future PR at some point, but atm the mod is in a much better position performance wise tha it was before afaik, and unless performance becomes a much greater priority, some of the more experimental/extreme optimisations in this PR may not be fully nessacery or at least as important ATM, at least rn.

PR #215, opened 2023-04-10
(Possibly Invalid) - Fix V-Sync with Nvidia when in exclusive FullScreen

thr3343:
This is a very simple/small PR that seems to fix an odd issue specific to Nvidia cards where in exclusive fullscreen, the frametime is extremely poor and unstable, and is practically unplayable unless exiting fullscreen or uncapping/disabling V-Sync completely.

This could possibly be due to my setup/Buggy Nvidia driver version, however decided to release it just in case it actually helps anyone else in anyway. (Or gives an actual/tangible performance/stability improvement)

### DIFF
--- a/src/main/java/net/vulkanmod/vulkan/Vulkan.java
+++ b/src/main/java/net/vulkanmod/vulkan/Vulkan.java
@@ -181,7 +181,7 @@ public class Vulkan {
 
     private static long allocator;
 
-    private static boolean vsync = false;
+    public static boolean vsync = false;
 
     private static StagingBuffer[] stagingBuffers;
     private static MemoryManager memoryManager;
--- a/src/main/java/net/vulkanmod/vulkan/shader/Pipeline.java
+++ b/src/main/java/net/vulkanmod/vulkan/shader/Pipeline.java
@@ -580,12 +580,13 @@ public class Pipeline {
                     ++i;
                 }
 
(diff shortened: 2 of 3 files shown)


thr3343:
https://user-images.githubusercontent.com/125277899/231013825-ae0a349d-94c3-4d17-bf67-8c97b9da9b2d.mp4

Applying this change seems to greatly improve stability and overall frametime (At least with my setup)

https://user-images.githubusercontent.com/125277899/231013943-f6824d4d-f66f-4d2b-afb1-68a18ac6c176.mp4

thr3343:
Also uploaded this just in case it actually fixes Broken Swapchain issues with RADV Drivers (and Possibly Nvidia Drivers as well) on Linux, however that is unlikely as they are most likely completely separate bugs, however want to rule out if they are in fact related or not

thr3343:
Eventually found that this is causes by a posible hardware flaw/Glitch/bug with Pascal where updating the UBO descriptors seem to cause the Graphics Queue to Stall in certain instances when V-Sync is enabled, causing the stutter.

This has a few possible fixes available, including manipulating the Sampler update order/Barriers and adding a fence to force a wait before the updated/rebinding of the descriptors, or using a Dedicated transfer queue.

However this doesn't seem to be a common problem and therefore may not necessarily warrant a direct fix, so will likely close the PR.

Issue #223, opened 2023-04-15
Linux arm64 support


Lolothepro:

bagong:
I would love to see this run on the Raspberry Pi 4 and on the Orange Pi 5, or other SBCs using  RK 3588(s) with the Mali GP-610 GPU. The later too are continuoisly getting better 3d  and Vulkan support, not only on Android, but also on Linux, and they are getting more and more affordable.

Lolothepro:
> I would love to see this run on the Raspberry Pi 4 and on the Orange Pi 5, or other SBCs using RK 3588(s) with the Mali GP-610 GPU. The later too are continuoisly getting better 3d and Vulkan support, not only on Android, but also on Linux, and they are getting more and more affordable.

https://gitlab.freedesktop.org/mesa/mesa/-/issues/8889

bagong:
So it's in the works? That's great to hear! Thanks!

vadcx:
@Lolothepro The bug report is pretty much non-actionable for a mesa developer:

- no version info of MC / VulkanMod specified to test
- devs with the Broadcom GPU are unlikely to have Minecraft (or if there's a working demo version with mod support, it should be specified)
- instructions how to install all that, including mods
- there's a stacktrace going into libvulkan_broadcom.so, but I think it's useless because the compiled library is probably specific to the distro. Then there are no debugging symbols to find out what this address translates to in source code

```
Stack: [0x0000fffefa7e0000,0x0000fffefafdf000],  sp=0x0000fffefafdd5a0,  free space=8181k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libvulkan_broadcom.so+0x6aa9c]
C  [libvulkan_broadcom.so+0x6b4d8]
C  [libvulkan_broadcom.so+0x850d4]
C  [libvulkan_broadcom.so+0x853a4]
C  [libvulkan_broadcom.so+0xf8bdc]
C  [libc.so.6+0x7d5c8]
```

Ideally you would want to do the entire testing and tracing back to code on your end, if you want to have a chance of any mesa contributor looking at it. This means (generally) recompiling mesa to some kind of a debug build, seeing if it still crashes when running it and checking the new crash address against the debug symbols that you have from the building process. This should conclusively point at the code that causes the crash.

You should be able to receive help with towards this kind of a report on their mailing list https://docs.mesa3d.org/lists.html

vadcx:
[...]

PR #245, opened 2023-05-24
add arm64/aarch64 support on linux 

leonpano2006:
if don't want to add arm64 support then please commant out


leonpano2006:
can fix this #223

PR #285, opened 2023-08-25
Add Driver and Vulkan version fields to the Debug Screen

thr3343:
This is a quick PR that allows the GPU Driver and Vulkan version to be decoded from binary format into a human readable version number, which is visible at any time when using the Debug Screen with F3.

This supports displaying the Driver versions for AMD, Nvidia and Intel. Other vendors may also work, but are not 100% guaranteed to display the driver version correctly.

Below example showing how it appears with a GTX 1080 Ti running Driver version 473.64.


thr3343:
Accidentally assigned the PR to 1.19.2, should be on 1.20, (My Mistake)

PR #295, opened 2023-09-12
Try out Architectury

CADIndie:
Adds Architectury and reformats VulkanMod to work with it, also includes some general cleanups.


CADIndie:
I should note that there are a few small things I need to fix like Javax Nullable being replaced with Jetbrains Nullable, if we choose to go this route I will make the changes.

PR #299, opened 2023-09-18
Use IDENTITY preTransform instead of currentTransform

artdeell:
This fixes the issue with the output being rotated when using SurfaceView on Android.
Before:
Screenshot_20230918-114259_PojavLauncher (Minecraft_ Java Edition для Android)
After:
Screenshot_20230918-114401_PojavLauncher (Minecraft_ Java Edition для Android)


artdeell:
Nevermind. This will need more changes as this onlty affects Android and may cause `VK_SUBOPTIMAL_KHR` errors.

PR #302, opened 2023-09-19
Feat: Implement surface pre-rotation

artdeell:
This is supposed to make application rendering more efficient by forcing the application to create a swapchain that is pre-rotated to the device's native screen orientation. It is necessary on Android devices.


artdeell:
Nvm i discovered a funky bug with block outlines

PR #305, opened 2023-09-23
Rewrite the optimisations to cubes to be more compatible with mods th…

Sollace:
Refiling of #284 (rebased to the new branch)

…at use custom cubes.

Also extends the feature to include normals, because why not? (note some cube implementations may have multiple faces with the same normal

The main approach I took was instead of creating a parallel set of vertices, I've rewritten it to instead collect the vertices from the vanilla cube right before rendering them. I dedupe them using a set and store them in the CubeModel class (which is now just a container) to avoid having to do the collecting again.

The transformVertices method works exactly like before, just with some changes to bring normal vectors into the fold.

It shouldn't perform any worse than before, though if it does become a factor the method can be rewritten to not use streams with relative ease. Since it's only done once per cube for the lifetime of the game, I don't think that will be neccessary.

This PR fixes compatibility with Mine Little Pony, as well as any other mod which uses Mson, or a similar library like it that attempts to use model containing cubes with customised vertices.

### DIFF
--- a/src/main/java/net/vulkanmod/interfaces/ModelPartCubeMixed.java
+++ b/src/main/java/net/vulkanmod/interfaces/ModelPartCubeMixed.java
@@ -1,8 +1,11 @@
 package net.vulkanmod.interfaces;
 
+import net.minecraft.client.model.geom.ModelPart.Polygon;
 import net.vulkanmod.render.model.CubeModel;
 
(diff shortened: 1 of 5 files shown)


Sollace:
Closing as there seems to be no interesting in merging this

PR #322, opened 2023-11-01
Remove macOS known issue

iMonZ:
Tested it on macOS.
Works perfectly, and it even detects my M1 GPU on macOS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ Demostration Video:
 
 ## Known Issues
 
-- Linux and macOS may not work.
 
 ## Notes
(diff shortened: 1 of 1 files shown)


johalun:
Tested 1.20.1 on an M3 Max Macbook. 2x the performance compared to Sodium, nice job!! Excited about the future of this project. 32 render distance, 5k resolution, 300+ fps.

xCollateral:
Thanks for letting us know. I'll close this PR since #383 exists and it is more extensive.

PR #309, opened 2023-09-25
Add Native Wayland Support + MailBox Mode

thr3343:
Rebase of https://github.com/xCollateral/VulkanMod/pull/291 onto the new 1.20.x default branch

This is mostly identical to the prior PR, outside of expanding Mailbox mode to be usable on all platforms, not just Wayland
(if the GPU driver supports MailBox Mode)

As before this won't include support for tearing-control-unstable-v1, (can't figure out how to enable/use it properly)
so Mailbox instead of Immediate mode will be used as the default display mode when Vsync is disabled

This also will not use swapchain images above 2 if Wayland + AMD is detected to avoid game breaking glitches
~~(Due to a suspected RADV driver bug)~~ 
(This was actually due to a limitation with the current Swapchain system used in the mod, 
this has been fixed and will be added when the new version of this PR is opened)

PR #325, opened 2023-11-03
Add Native Wayland Support V2 + Minor Nvidia/Performance Optimisations

thr3343:
This is similar to the [prior PR](https://github.com/xCollateral/VulkanMod/pull/309), but with some various technical enhancements over the original

* Use Alternate Swapchain System
    * Designed to fix and address major gamebreaking stability issues with the Mesa/RADV driver on Linux
    * Should not cause any issues on other platforms AFAIK
    * Swapchain is adjusted based on GPU minImageCount
        * Which may improve performance on some systems
    
    
  
* Much more reliable display server detection
    * The original PR used a flawed detection method, that often failed to detect X11/Wayland properly
 
* FastSync (aka Mailbox Mode) is used by default for VSync Off
    * Falls back to Immediate mode if not supported
    * This change was made to allow exposing FastSync on all platforms if supported
        * (instead of being restricted to Wayland only)
    * Also might be needed for FreeSync (Unconfirmed, i'm probably wrong)

* Minor performance optimisations
    * includes a Nvidia specific adjustment to use 24-bit depth format instead of 32-bit
    * This also does some minor code cleanup, but have no issue reverting these if needed

### DIFF
--- a/src/main/java/net/vulkanmod/config/Config.java
+++ b/src/main/java/net/vulkanmod/config/Config.java
@@ -13,6 +13,7 @@ import java.util.Collections;
 public class Config {
 
     public int frameQueueSize = 2;
+    public int minImageCount = 3;
     public VideoResolution resolution = VideoResolution.getFirstAvailable();
     public boolean windowedFullscreen = false;
     public boolean guiOptimizations = false;
--- a/src/main/java/net/vulkanmod/config/Config.java
+++ b/src/main/java/net/vulkanmod/config/Config.java
@@ -13,7 +13,6 @@ import java.util.Collections;
 public class Config {
 
     public int frameQueueSize = 2;
-    public int minImageCount = 3;
     public VideoResolution resolution = VideoResolution.getFirstAvailable();
     public boolean windowedFullscreen = false;
     public boolean guiOptimizations = false;
--- a/src/main/java/net/vulkanmod/config/Options.java
+++ b/src/main/java/net/vulkanmod/config/Options.java
@@ -16,8 +16,6 @@ public class Options {
     static Window window = Minecraft.getInstance().getWindow();
     public static boolean fullscreenDirty = false;
 
-
-
     public static Option<?>[] getVideoOpts() {
         return new Option[] {
                 new CyclingOption<>("Resolution",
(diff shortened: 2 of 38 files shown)


ElectricLizzzard:
I cannot break it in any way. It passed all stress tests and tested in real life scenarios with just playing it. Rock stable. Platform detection also works as expected.

Etaash-mathamsetty:
tysm for the RADV fixes, finally somebody figured out what the issue was after 2 years
```
Swapchain is adjusted based on GPU minImageCount

    Which may improve performance on some systems
```
I tried to make a PR for this and it never got merged lol, even though the code back then wasn't following spec

Etaash-mathamsetty:
works as expected and working better than the main branch is on RADV, tysm once again

thr3343:
Shouldn't have done this, but have added some Chunk rendering-related optimisations i had lying around

This is bad practice as this commit should be in a separate PR entirely.
But have been sitting on these chunk code optimisations for months, so got too impatient and added them anyway

Will make sure not to add any more commits to avoid this PR getting too bloated (Outside of emergency fixes/game-breaking bugs)

thr3343:
Apologies for the huge [Use Queue Enum](https://github.com/xCollateral/VulkanMod/pull/325/commits/8b589c7dcfc306641824017cca921dffc69ca3fd) commit
It was originally for a refactor PR I was planning to do later in the year, but i misclicked and added it by complete accident.

It has a very minor FPS improvement and it also fixes VSync on old Nvidia GPUs, 
But outside that it is alot of refactoring for not much benefit, and it has nothing to do with Wayland at all.
So feel free to remove it if needed.

ElectricLizzzard:
I would state that right now this PR is giving seamless from Vanilla experience both on Wayland and X11. The only one difference is monstrous amount of frames per second. The only times my GPU not hitting 100% load is in locations with a large number of entities, but this is completely different story.

There is only problem with Nvidia being BAD on Wayland. I hope this will be fixed soon by Nvidia itself.

Last bug that do not affect gameplay in any way but still irritates is swapchain destroy bug, when game is closed by user and this cause instance lag and crash with error 6. But this is not that's PR explicit bug. It appears in stable release too.

thr3343:
Reverted alot of the performance optimisations as one of them was causing specific errors on Staging buffer copies, and due to this PR being a mess the breaking commit couldn't be tracked down.

So used a Squash to delete/nuke the commits, including the breaking change(s).

Don't mind doing this as I have the unedited original + commits in a separate "backed up" branch made yesterday.

_(So nothing was lost anyway TBH)_

thr3343:
Despite what I said above, i did decide in the end to re-add some of the optimisation-related changes,
These are relatively minor code changes unlike before, so the chance/risk of subsequent crashes/issues is very low, at least AFAIK

PR #328, opened 2023-11-19
Chunk Rendering Optimisations + VRAM Usage Reductions

thr3343:
This was originally a part of an older version of the Wayland PR,
But was later spun off into a separate PR with additional improvements and cleanups

This heavily leverages and exploits TerrainRenderTypes to enable certain optimisations, 
This will replace much of the prior usages of RenderTypes, so hopefully this won't cause too many issues

Quick/Brief Summary _(Badly formatted sorry)_
* Greatly reduce CPU overhead of PushConstants/Chunk Translations with Draw Direct
  * PushConstants are now updated per ChunkArea instead of per DrawCall
  * Can reduce CPU overhead by over 100x depending on number of DrawCommands per ChunkArea
 * Merge Direct and Indirect shaders into one unified Terrain shader
* Refactor DrawBuffers structure to improve CPU Cache Locality
  * (i.e. Using Structure of Arrays (SoA) + reducing class sizes in hot code paths)
* Greatly reduce rate of empty Draw calls
  *   Usually only very rarely occur when new chunk generation is happening
    * _(All other types of conventional Chunk Loads no longer generate empty Draw calls AFAIK)_
* Potentially make Chunk Build Tasks slightly faster
*  Use per TerrainRenderType AreaBuffers to improve GPU Utilisation
* Selectively enable/Allocate AreaBuffers based on needed TerrainRenderType
  * This helps to reduce VRAM utilisation, especially if chunks don't used mixed Types
  * (i.e. If the chunk(s) are only solid or only translucent materials)


thr3343:
Think i'll do a way better + cleaned up version of this PR rebased on 0.3.6 code, so will close this atm.

PR #337, opened 2023-12-05
Add Experimental GPU Selector

thr3343:
This is a quick PR to fix an issue with some multi-GPU users, where they are unable to select their preferred GPU due to GPU ordering priority

This adds the ability to switch between auto mode, which is the current system, and a preferred GPU override.
So under default conditions the Initialisation process is mostly unchanged

This PR also adds some non critical functionality such more detailed Log output, including the GPU being selected, and  GPU error/failure reasons.

### DIFF
--- a/src/main/java/net/vulkanmod/config/Config.java
+++ b/src/main/java/net/vulkanmod/config/Config.java
@@ -20,6 +20,7 @@ public class Config {
     public boolean indirectDraw = false;
     public boolean uniqueOpaqueLayer = true;
     public boolean entityCulling = true;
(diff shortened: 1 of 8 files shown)


ElectricLizzzard:
This feature is very needed as i saw in community.

thr3343:
TBH i had alot of this code lying around already done.
So we agreed that i could open this PR so it could be used as a reference to help with Development.
Its drastically improved over this PR, so will be closing this as it has achieved its purpose and what it needed to do tbh.

PR #338, opened 2023-12-08
Add BaseInstance Translations & per ChunkArea PushConstants + additional VRAM reductions and CPU-related tweaks

thr3343:
This PR adds some leftover optimisations from an older version of the Wayland PR, namely CPU overhead improvements pertaining to Chunk rendering, and some minor Shader changes

Also added some Optifine Inspired Options: Animations and Render Sky:
Which toggle both Animations and Sky Rendering Respectively

*Quick Summary of Changes*

* Make PushConstants per ChunkArea instead of per Drawcall 

* Use BaseInstance to handle per draw Translations instead
  * Massively reduces CPU overhead, especially with high Drawcall dense scenes/Large Render Distances 

* Add Early-Z version of the terrain Shader, 
  * This enables Early fragment Tests, which may give a slight FPS improvement on some Devices
  * Only effects the Translucent layer currently

* Misc. Class + Record refactoring to improve GC + JIT efficiency 
  * Hopefully should also be more CPU cache friendly + improve Cache locality

* Combine terrain_direct and terrain_indirect into one unified Shader

* Reduce CPU overhead of Sky blending

* Add Disable Animations and RenderSky Options


src/main/resources/assets/vulkanmod/shaders/basic/terrain/terrain_Z.fsh
thr3343, line 1: These are potentially very risky shader optimisations;
But as far as I can tell this didn't seem to change or effect rendering at all, and didn't seem to cause any tangible issues or graphical artefacts of any kind

Ideally would like to keep these as it noticeably improves fragment/rasterisation perf
But have no issue whatsoever if these need to be removed for any reason 
_(e.g. Someone finds a graphical artefact/glitch this directly causes)_

thr3343:
Will close this and redo a new version;  has major performance regressions due to the
[Improve Empty drawParameter Handling](https://github.com/xCollateral/VulkanMod/pull/338/commits/1a0cf131bacbf2fbc6378695d934479d3deec3fa) Commit

PR #345, opened 2023-12-16
Per ChunkArea PushConstants + VRAM Usage Reductions

thr3343:


thr3343:
Thanks to Collateral, they pointed out that this PR has major issues with CPU cache Misses
Which causes performance  regressions with the BFS algorithm used to handle culling/uploads during chunk loads.
Causing Chunk loads to become alot slower than normal

i.e. I had hyperfixated on reducing class sizes to improve cache line alignment, but completely overlooked about cache misses

```
if(!renderSection.isCompletelyEmpty()) {
//This code has major issues with cache misses due to the EnumMap in DrawBuffers afaik
                for(var t : renderSection.getCompiledSection().renderTypes) {
                    drawBuffers.addDrawCommands(t, renderSection.getDrawParameters(t));
                }
                this.drawBufferSetQueue.add(drawBuffers);
                this.nonEmptyChunks++;
}
```
Perf regressions are not ideal as over time + PRs, they can accumulate and become hard to debug
So atm I decided its best this PR will be left to a later version 
So these cache miss issues can be ironed out TBH

PR #355, opened 2024-01-08
Add support for asdf

zastrixarundell:
Add `.tool-versions` file to support asdf to install the correct version of java and gradle. With this users only need to run:

* asdf plugin add java
* asdf plugin add gradle
* asdf install
* gradle build

Helps a lot when someone needs to do a quick build.


xCollateral:
I don't think this PR is really needed.

PR #358, opened 2024-01-09
Per ChunkArea PushConstants + VRAM Usage Reductions V2

thr3343:
Similar to the prior PR: https://github.com/xCollateral/VulkanMod/pull/345
But with major code cleanups + some additional enhancements and improvements:

* In-game renderFog Spec Constant
  * Spec Constants allows for adaptive shader optimizations to be applied by the Driver
  * Has the advantage of allowing performance improvements, instead of just setting the fog uniforms to zero
* CPU cache-miss fixes
  * Prior PR had a major problem with high CPU cache misses, causing major chunk loading lag and regressions
  * This has been fixed so chunk loads perform roughly the same as upstream
 * Render pass optimizations + mode Switching
   * `DefaultPass` is used instead of `LegacyMainPass` when no Post Effects are active to improve performance 
   * `DefaultPass` also merged into `LegacyMainPass` to allow for static final inlining + more concise code
  * Color clear skips
    * Color Attachment clears are optimized out + skipped if the current clear color has already been applied 
    * This helps majorly reduce pipeline stalls, which can greatly improve performance at high framerates
 * Low VRAM mode
   * Splits the vertex Areabuffer into separate per Rendertype vertexbuffers
   * Allows for more granular + precise resizing, which can reduce VRAM by roughly 20% or more
   * This can effect performance on some hardware, so its disabled by default

- Issues/Bugs: 
- Glowing Effect is force disabled 
  - Glowing is broken with RenderPass mode switching due to missing Drawcalls as a result of them getting assigned to the wrong RenderPass, 
    - _(i.e. is drawn to a Color Attachment instead of the swapchain)_
  - This causes flickering artefacts for a single frame when enabling/disabling Glowing
  - Which is not ideal and effects gameplay 
  - So atm this is disabled until i find a better fix tbh
- All other Shader effects are functional however: as this issue applies solely to Glowing


src/main/java/net/vulkanmod/render/chunk/DrawBuffers.java
```diff
@@ -200,73 +242,24 @@ private static void fakeIndirectCmd(VkCommandBuffer commandBuffer, IndirectBuffe
         }
     }
 
-    public void buildDrawBatchesDirect(StaticQueue<RenderSection> queue, Pipeline pipeline, RenderType renderType, double camX, double camY, double camZ) {
-        TerrainRenderType terrainRenderType = TerrainRenderType.get(renderType);
-        terrainRenderType.setCutoutUniform();
+    public void buildDrawBatchesDirect(StaticQueue<DrawParameters> queue, TerrainRenderType terrainRenderType, double camX, double camY, double camZ, long layout) {
+        if(queue==null || queue.size() == 0) return;
```
zastrixarundell: You *could* use an [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) here .

thr3343: Thanks for the suggestion: Think i'l just do a minor draw code cleanup commit instead though if that's OK


(There's some duplicate code which can be extracted as a common part: this will use an AND condition which won't work as an Optional afaik)

```c++
if(typedSectionQueue!=null && typedSectionQueue.size() != 0) {
    chunkArea.drawBuffers().bindBuffers(terrainRenderType, commandBuffer, camX, camY, camZ, layout);
    if (indirectDraw) chunkArea.drawBuffers().buildDrawBatchesIndirect(indirectBuffers[currentFrame], typedSectionQueue, terrainRenderType);
    else chunkArea.drawBuffers().buildDrawBatchesDirect(typedSectionQueue, terrainRenderType);
}
```

thr3343:
Realised its actually better after all to split this into some separate PRs w/ minimal merge conflicts, 
so will close this rn and reopen this later

PR #360, opened 2024-01-12
Add renderFog Spec Constant + Animations and renderSky options

thr3343:
Adds basic Graphics options inspired by Optifine, namely Animation, Sky and Render Fog toggles

The RenderFog Toggle uses a SPIR-V exclusive feature called Specalization Constants, which enables dynamically modifying shader code during runtime without Shader recompilations

Allowing to the ability to disable specific effects whilst being able to apply shader optimisations at the same time


thr3343:
Outdated due to upstream changing significantly since this was last updated: Will close and reopen as an updated branch/PR.

PR #363, opened 2024-01-13
Add alternative Queue enum

thr3343:
Adds an alternative Queue enum Structure, which consolidates the current Queues into one singular enum, allowing the other Queue Classes to be removed

This is primarily intended to allow easy switching between Queues for easy debugging troublehsooting in cases 

This PR also adds some additional queue functions, namely TrimCmdPool, FillBuffer, and Gigabarriers


thr3343:
Replaced with this PR: https://github.com/xCollateral/VulkanMod/pull/393

PR #370, opened 2024-02-02
Fix custom shader NPE

### LINKED ISSUE #364: Shaders cannot be compiled
I am building a MMD model（a type of high-precision model） renderer, when I am writing shaders for resolving performance issues, I found it cannot be compiled on vulkan renderer and there aren't any error logs available.  I cannot prevent crashing using NEC mod, because after the shader null pointer exception is a native exception. 
loader source: https://github.com/PrimogemStudio/Advanced-Framework
the shader is at src/resources/assets/advancedfmkshaders/core/<any files start with mmd_entity>
refrences are at src/main/java/com/primogemstudio/advancedfmk/mmd/renderer/Renderer.kt
image
(idea debugger, the model has 68604 vertices, throws NullPointerException)
[hs_err_pid10184.log](https://github.com/xCollateral/VulkanMod/files/13933849/hs_err_pid10184.log)
And it causes the native exception and the whole JVM program crashes.

MetsukiMio:
Close #364 
[crash-2024-02-02_08.40.32-client.txt](https://github.com/xCollateral/VulkanMod/files/14138373/crash-2024-02-02_08.40.32-client.txt)
@thr3343 After debugging, I found that the real exception happens in this mixin, it had been hard coded and cannot build the shader path correctly.

### DIFF
--- a/src/main/java/net/vulkanmod/mixin/render/ShaderInstanceM.java
+++ b/src/main/java/net/vulkanmod/mixin/render/ShaderInstanceM.java
@@ -73,7 +73,9 @@ public class ShaderInstanceM implements ShaderMixed {
 
         try {
             if(Pipeline.class.getResourceAsStream(String.format("/assets/vulkanmod/shaders/minecraft/core/%s/%s.json", name, name)) == null) {
-                createLegacyShader(resourceProvider, new ResourceLocation("shaders/core/" + name + ".json"), format);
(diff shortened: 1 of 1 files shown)


xCollateral:
This issue should be fixed now. Even Fabric API shader loading has been implemented, which should help with custom shader loading.

PR #373, opened 2024-02-10
Fix _getTexLevelParameter returning -1 for everything

Sollace:
Changed `target == GL11.GL_TEXTURE_2D` to `target != GL11.GL_TEXTURE_2D`.

This fixes compatibility with Mine Little Pony and possibly other mods that make use of GlStateManager._getTexLevelParameter


xCollateral:
It should be solved.

PR #386, opened 2024-03-14
[CPU/GPU Overhead] Optimize samplers and Core Shaders + Skip Pipeline/Descriptorset Binds and reduce uniform memory usage

thr3343:
This PR consits of two parts: Bind skip and Minor Sampler optimisations

#### Bind Skip
Reduces vkBindPipeline and vkBindDescriptorSets calls by Approx. 50%, including reducing DescriptorSet updates by 40% and Uniform memory usage by Approx. 75%

Also includes some Misc. + minor performance optimisations, which were included to help improve the effectiveness of vkBindPipeline skips

#### Minor Sampler optimisations
The Sampler optimisations reworks sampler handling slightly to allow more precise stage flags, allowing the GPU driver to potential apply additional optimisations and boost performance slightly

#### Shader Refactors
Removed many vanilla core shaders to reduce required resources + facilitate pipeline reuse
Also applied some minor optimisations, including removing unused/unnecessary uniforms

#### Bugs

The title screen appears slightly more aliased due to lack of blurring, but this is the only visual bug with this PR afaik

PR #392, opened 2024-04-04
[CPU/GPU Tweaks] Misc. GPU+CPU performance fixes +  Optimise Core Shaders

thr3343:
This is based from the prior PR: https://github.com/xCollateral/VulkanMod/pull/386

But with some additional performance-related patches + critical rendering fixes:

* Includes Color Attachment Clear Skips to reduce pipeline stalls, and CPU optimisations for Draw Direct

* The Uniform memory reduction and Mob shader optimisations are removed to fix several visual issues
  * This will likely regress performance slightly, but in exchange should improve stability.


thr3343:
Will replace this with a improved version without performance hacks using Bindless rendering techniques from Vulkan 1.2 _(Mostly Descriptor Indexing)_

PR #393, opened 2024-04-06
[Chunk Uploads + Enum Refactors] Improve Chunk upload speed + Refactor MemoryTypes and VkQueues

thr3343:
Cleans up significant boilerplate in several classes related to MemoryTypes + Queues, and replaces them with a much more readable and easier to use Enum-based system

This PR also corrects a mistake where cached system RAM was used instead of non-cached VRAM for Chunk uploads.
Which slightly improves the speed of chunk loading and saving.


thr3343:
Decided to close this as its outdated + causes crashes on Non-Resizable BAR GPUs if more than 256MB needs to be uploaded in a single frame _(e.g. high res texture Packs)_

It also wasn't clear if this PR actually improved Chunk load speed, so will revisit + overhaul this code at a later time most likely

PR #383, opened 2024-03-01
Rework of the Readme.md

SpidFightFR:
Just a little rework of the welcome page.

Feel free to bring suggestions or edits ! :smile: :+1:

PR #400, opened 2024-04-08
updated README file

lvllaby:


Felix14-v2:
Why should they change it?

lvllaby:
on their site, they have a link to modrinth and not curseforge

Enderteck:
You should still put both I think but I'm all for it (modrinth is better)

Enderteck:
You should also remove Linux from this note as it's supported now. MacOS is complicated though so you should leave it there.

README.md:

> Known Issues
> Linux and macOS may not work.
> 

Here's why it's complicated with apple (tldr: they're assholes like ususal):

https://en.wikipedia.org/wiki/Vulkan#Apple

lvllaby:
added modrinth button and change linux note

Zorklis:
"Linux is now supported" should be in the notes and not in Known Issues then.
Might as well add windows and change it to: "Windows and Linux are the supported platforms" just to clear the air

xCollateral:
Duplicate of #383

PR #412, opened 2024-05-03
Rendering-related fixes, clean-up and DEBUG_LINES support

VisualEhrmanntraut:
This fixes mods the likes of Ad Astra amongst other problems.

image

image

### DIFF
--- a/src/main/java/net/vulkanmod/vulkan/Drawer.java
+++ b/src/main/java/net/vulkanmod/vulkan/Drawer.java
@@ -2,6 +2,7 @@ package net.vulkanmod.vulkan;
 
 import com.mojang.blaze3d.vertex.VertexFormat;
 import net.vulkanmod.vulkan.memory.*;
+import net.vulkanmod.vulkan.memory.AutoIndexBuffer.DrawType;
 import net.vulkanmod.vulkan.util.VUtil;
 import org.lwjgl.system.MemoryUtil;
 import org.lwjgl.vulkan.VkCommandBuffer;
@@ -25,9 +26,10 @@ public class Drawer {
     private int framesNum;
     private VertexBuffer[] vertexBuffers;
     private final AutoIndexBuffer quadsIndexBuffer;
-    private final AutoIndexBuffer triangleFanIndexBuffer;
-    private final AutoIndexBuffer triangleStripIndexBuffer;
     private final AutoIndexBuffer linesIndexBuffer;
+    private final AutoIndexBuffer triangleFanIndexBuffer;
+    private final AutoIndexBuffer stripIndexBuffer;
+    private final AutoIndexBuffer debugLinesIndexBuffer;
     private final AutoIndexBuffer lineStripIndexBuffer;
     private UniformBuffer[] uniformBuffers;
 
@@ -36,9 +38,10 @@ public class Drawer {
     public Drawer() {
         //Index buffers
         quadsIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.QUADS);
-        triangleFanIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.TRIANGLE_FAN);
-        triangleStripIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.TRIANGLE_STRIP);
         linesIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.LINES);
+        triangleFanIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.TRIANGLE_FAN);
+        stripIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.TRIANGLE_STRIP);
+        debugLinesIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.DEBUG_LINES);
         lineStripIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.LINE_STRIP);
     }
 
@@ -79,8 +82,12 @@ public class Drawer {
         vertexBuffer.copyToVertexBuffer(vertexFormat.getVertexSize(), vertexCount, buffer);
 
         switch (mode) {
-            case QUADS, LINES -> {
+            case QUADS -> {
                 autoIndexBuffer = this.quadsIndexBuffer;
+                indexCount = vertexCount / 4 * 6;
+            }
+            case LINES -> {
+                autoIndexBuffer = this.linesIndexBuffer;
                 indexCount = vertexCount / 4 * 6;
             }
             case TRIANGLE_FAN -> {
@@ -88,11 +95,11 @@ public class Drawer {
                 indexCount = (vertexCount - 2) * 3;
             }
             case TRIANGLE_STRIP, LINE_STRIP -> {
-                autoIndexBuffer = this.triangleStripIndexBuffer;
+                autoIndexBuffer = this.stripIndexBuffer;
                 indexCount = vertexCount;
             }
             case DEBUG_LINES -> {
-                autoIndexBuffer = this.linesIndexBuffer;
+                autoIndexBuffer = this.debugLinesIndexBuffer;
                 indexCount = vertexCount;
             }
             case DEBUG_LINE_STRIP -> {
@@ -112,19 +119,23 @@ public class Drawer {
     }
 
     public AutoIndexBuffer getQuadsIndexBuffer() {
-        return quadsIndexBuffer;
-    }
-
-    public AutoIndexBuffer getTriangleFanIndexBuffer() {
-        return triangleFanIndexBuffer;
-    }
-
-    public AutoIndexBuffer getTriangleStripIndexBuffer() {
-        return triangleStripIndexBuffer;
+        return this.quadsIndexBuffer;
     }
 
     public AutoIndexBuffer getLinesIndexBuffer() {
         return this.linesIndexBuffer;
+    }
+
+    public AutoIndexBuffer getTriangleFanIndexBuffer() {
+        return this.triangleFanIndexBuffer;
+    }
+
+    public AutoIndexBuffer getStripIndexBuffer() {
+        return this.stripIndexBuffer;
+    }
+
+    public AutoIndexBuffer getDebugLinesIndexBuffer() {
+        return this.debugLinesIndexBuffer;
     }
 
     public AutoIndexBuffer getLineStripIndexBuffer() {
@@ -156,13 +167,15 @@ public class Drawer {
         vkCmdDraw(commandBuffer, vertexCount, 1, 0, 0);
     }
 
-    public void bindAutoIndexBuffer(VkCommandBuffer commandBuffer, int drawMode) {
+    public void bindAutoIndexBuffer(VkCommandBuffer commandBuffer, DrawType drawMode) {
         AutoIndexBuffer autoIndexBuffer;
         switch (drawMode) {
-            case 7 -> autoIndexBuffer = this.quadsIndexBuffer;
-            case 6 -> autoIndexBuffer = this.triangleFanIndexBuffer;
-            case 5 -> autoIndexBuffer = this.triangleStripIndexBuffer;
-            default -> throw new RuntimeException("unknown drawType");
+            case QUADS -> autoIndexBuffer = this.quadsIndexBuffer;
+            case TRIANGLE_FAN -> autoIndexBuffer = this.triangleFanIndexBuffer;
+            case LINE_STRIP, TRIANGLE_STRIP -> autoIndexBuffer = this.stripIndexBuffer;
+            case DEBUG_LINES -> autoIndexBuffer = this.debugLinesIndexBuffer;
+            case LINES -> autoIndexBuffer = this.linesIndexBuffer;
+            default -> throw new IllegalArgumentException(String.format("Unexpected value: %s", drawMode));
         }
         IndexBuffer indexBuffer = autoIndexBuffer.getIndexBuffer();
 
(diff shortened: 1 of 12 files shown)


thr3343:
TBH I was also going to open a PR on this issue
But this adds more draw modes, while my version only added Lines, so this is the better PR imho

Will link my branch anyway, just in case it is useful as a reference: https://github.com/thr3343/VulkanMod/tree/DrawMode

thr3343 (changes requested):
_(I'm bad with Git, apologies for the formatting)_

Made an attempt to fix MiniHUD rendering by adjusting Strip vertex count and moving DEBUG_LINES to Quads.

If this patch doesn't work/isn't effective, i'm happy to mark this review as outdated.

```diff
Subject: [PATCH] Change DEBUG_LINES to Quads
---
Index: src/main/java/net/vulkanmod/render/VBO.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/net/vulkanmod/render/VBO.java b/src/main/java/net/vulkanmod/render/VBO.java
--- a/src/main/java/net/vulkanmod/render/VBO.java	(revision 76101bffa4dbb94fd1300ad202013e76e8412954)
+++ b/src/main/java/net/vulkanmod/render/VBO.java	(revision 0b1a7f8d4700bf460afd758edfe255b52dfef003)
@@ -13,9 +13,13 @@
 import net.vulkanmod.vulkan.shader.GraphicsPipeline;
 import net.vulkanmod.vulkan.shader.Pipeline;
 import org.joml.Matrix4f;
+import org.lwjgl.opengl.GL11;
 
 import java.nio.ByteBuffer;
 
+import static org.lwjgl.vulkan.VK10.*;
+import static org.lwjgl.vulkan.VK10.VK_POLYGON_MODE_FILL;
+
 @Environment(EnvType.CLIENT)
 public class VBO {
     private VertexBuffer vertexBuffer;
@@ -67,7 +71,7 @@
                 }
                 case QUADS, LINES -> {
                     autoIndexBuffer = Renderer.getDrawer().getQuadsIndexBuffer();
-                    this.indexCount = vertexCount;
+                    this.indexCount = vertexCount / 4 * 6;
                 }
                 case TRIANGLES -> {
                     autoIndexBuffer = null;
@@ -124,7 +128,26 @@
             RenderSystem.assertOnRenderThread();
 
             VRenderSystem.applyMVP(MV, P);
-
+            switch (mode.asGLMode) {
+                case GL11.GL_LINES -> {
+                    VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_LINE_LIST;
+                    VRenderSystem.polygonMode = VK_POLYGON_MODE_LINE;
+                }
+                case GL11.GL_LINE_STRIP -> {
+                    VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_LINE_STRIP;
+                    VRenderSystem.polygonMode = VK_POLYGON_MODE_LINE;
+                }
+                // TODO: Use Triangle Fan topology if supported by the device
+                case GL11.GL_TRIANGLE_FAN, GL11.GL_TRIANGLES -> {
+                    VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
+                    VRenderSystem.polygonMode = VK_POLYGON_MODE_FILL;
+                }
+                case GL11.GL_TRIANGLE_STRIP -> {
+                    VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP;
+                    VRenderSystem.polygonMode = VK_POLYGON_MODE_FILL;
+                }
+                default -> throw new RuntimeException(String.format("Unknown VertexFormat mode: %s", mode));
+            }
             Renderer renderer = Renderer.getInstance();
             renderer.bindGraphicsPipeline(pipeline);
             renderer.uploadAndBindUBOs(pipeline);
Index: src/main/java/net/vulkanmod/vulkan/Drawer.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/net/vulkanmod/vulkan/Drawer.java b/src/main/java/net/vulkanmod/vulkan/Drawer.java
--- a/src/main/java/net/vulkanmod/vulkan/Drawer.java	(revision 76101bffa4dbb94fd1300ad202013e76e8412954)
+++ b/src/main/java/net/vulkanmod/vulkan/Drawer.java	(revision 0b1a7f8d4700bf460afd758edfe255b52dfef003)
@@ -89,7 +89,7 @@
             }
             case TRIANGLE_STRIP, LINE_STRIP -> {
                 autoIndexBuffer = this.triangleStripIndexBuffer;
-                indexCount = vertexCount;
+                indexCount = (vertexCount - 2) * 3;
             }
             case DEBUG_LINES -> {
                 autoIndexBuffer = this.linesIndexBuffer;
Index: src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java b/src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java
--- a/src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java	(revision 76101bffa4dbb94fd1300ad202013e76e8412954)
+++ b/src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java	(revision 0b1a7f8d4700bf460afd758edfe255b52dfef003)
@@ -26,11 +26,11 @@
                 size = vertexCount / 4 * 6 * IndexBuffer.IndexType.SHORT.size;
                 buffer = genQuadIndices(vertexCount);
             }
-            case TRIANGLE_FAN -> {
+            case TRIANGLE_FAN, TRIANGLE_STRIP -> {
                 size = (vertexCount - 2) * 3 * IndexBuffer.IndexType.SHORT.size;
                 buffer = genTriangleFanIndices(vertexCount);
             }
-            case TRIANGLE_STRIP, LINES, LINE_STRIP -> {
+            case LINES, LINE_STRIP -> {
                 size = vertexCount * IndexBuffer.IndexType.SHORT.size;
                 buffer = genIncrementalIndices(vertexCount);
             }
```

VisualEhrmanntraut:
MiniHUD works fine now

src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java
```diff
@@ -44,89 +48,78 @@ private void createIndexBuffer(int vertexCount) {
...
+            indices.put(j + 5, (short) i);
 
             j += 6;
         }
 
         return buffer;
-        //this.type.copyIndexBuffer(this, bufferSize, idxs);
     }
 
-    public static ByteBuffer genTriangleFanIdxs(int vertexCount) {
```
thr3343: The Skybox gradient doesn't render correctly in this PR
Issues like this are usually due to incorrect index ordering, so not a big deal to fix most likely

UpStream
b

This PR
a



Think the Skybox uses TriangleFan topology, which is why I put this comment here, but could easily be wrong about that though.

src/main/java/net/vulkanmod/vulkan/Drawer.java
```diff
@@ -25,17 +26,23 @@ public class Drawer {
...
+    private final AutoIndexBuffer stripIndexBuffer;
+    private final AutoIndexBuffer debugLinesIndexBuffer;
+    private final AutoIndexBuffer lineStripIndexBuffer;
     private UniformBuffer[] uniformBuffers;
 
     private int currentFrame;
 
     public Drawer() {
         //Index buffers
         quadsIndexBuffer = new AutoIndexBuffer(UINT16_INDEX_MAX, AutoIndexBuffer.DrawType.QUADS);
```
thr3343: _(This is complete nitpick, so feel free to skip this if you prefer)_

`UINT16_INDEX_MAX` should be 65536 instead of 98304: as Indices > 65536 overflow as a result of using `VK_INDEX_TYPE_UINT16`
_(Minecraft uses 16-bit instead of 32-Bit indices as they can have performance advantages)_

Which causes the index Buffers to be about 33% larger than needed.

This is my mistake as I added this `UINT16_INDEX_MAX` constant in an older VulkanMod version, but never corrected it, so that's MB.

Put a [small commit here](https://github.com/thr3343/VulkanMod/commit/730c66fe8d95958e66084a81393e29e76f4814db) w/ this optimization in case it helps as a reference.

src/main/java/net/vulkanmod/vulkan/memory/AutoIndexBuffer.java
VisualEhrmanntraut: Screenshot 2024-05-09 at 23 04 36
This looks fine? @thr3343

thr3343: Thanks for the info, afaict this is potentially a driver/hardware-specific issue, as I can reproduce it without fail on my Nvidia GPU

This is not the fault of the PR: Hardware specific bugs like this are the absolute worst to debug as they are impossible to reproduce without the right GPU.

There are some Nvidia users in the Discord, so sometime later will ask them to test this PR, to confirm if it's indeed Nvidia-specific or if its just an issue on my System.

x

VisualEhrmanntraut: It is not. I just made a typo and will commit the fix when possible.

src/main/java/net/vulkanmod/vulkan/Drawer.java
VisualEhrmanntraut: `65536 / 4 * 6 > 65536`?
edit: nvm, 6 hour sleep is breaking my logic

src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
```diff
@@ -163,6 +171,40 @@ public static void depthMask(boolean b) {
...
+            case GL11.GL_LINE_STRIP -> {
+                VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_LINE_STRIP;
+            }
+            // TODO: Use Triangle Fan topology if supported by the device
+            case GL11.GL_TRIANGLE_FAN, GL11.GL_TRIANGLES -> {
+                VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST;
+            }
+            case GL11.GL_TRIANGLE_STRIP -> {
+                VRenderSystem.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP;
+            }
```
xCollateral: I'm sure enough there's no need for `VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP` topology as it can be emulated by `VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST` and a properly set up Index Buffer, reducing the number of pipelines.
Same goes for` VK_PRIMITIVE_TOPOLOGY_LINE_STRIP` which could be expressed by `VK_PRIMITIVE_TOPOLOGY_LINE_LIST` + Index Buffer

src/main/java/net/vulkanmod/vulkan/shader/GraphicsPipeline.java
```diff
@@ -161,7 +156,7 @@ private long createGraphicsPipeline(PipelineState state) {
 
             VkPipelineDynamicStateCreateInfo dynamicStates = VkPipelineDynamicStateCreateInfo.calloc(stack);
             dynamicStates.sType(VK_STRUCTURE_TYPE_PIPELINE_DYNAMIC_STATE_CREATE_INFO);
-            dynamicStates.pDynamicStates(stack.ints(VK_DYNAMIC_STATE_DEPTH_BIAS, VK_DYNAMIC_STATE_VIEWPORT, VK_DYNAMIC_STATE_SCISSOR));
+            dynamicStates.pDynamicStates(stack.ints(VK_DYNAMIC_STATE_DEPTH_BIAS, VK_DYNAMIC_STATE_VIEWPORT, VK_DYNAMIC_STATE_SCISSOR, VK_DYNAMIC_STATE_LINE_WIDTH));
```
xCollateral: I suppose `VK_DYNAMIC_STATE_LINE_WIDTH` should only be added when using `VK_PRIMITIVE_TOPOLOGY_LINE_*`

xCollateral (changes requested):

src/main/java/net/vulkanmod/vulkan/shader/GraphicsPipeline.java
VisualEhrmanntraut: According to the docs, the only restriction with the wide line feature not being supported is that the line width cannot be anything other than 1.0.

VisualEhrmanntraut: (So I did it for simplicity)

src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
VisualEhrmanntraut: Is there a performance benefit to doing this? I did this because it's what Minecraft does.

xCollateral: Less pipelines and less state changes. Probably there is not much difference in most scenarios (not many topology changes). Though it's probably worth since it's not that big of a change.

xCollateral:
I wanted to thank you for your work, I'm interested in merging this PR as it resolves a compatibility issue I haven't got around to fix yet.

src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
VisualEhrmanntraut: So, you want me to change it back to emulating using triangle list, correct?

thr3343 (approved):

VisualEhrmanntraut:
LINES mode is still a bit fishy so don't merge it yet until we figure it out :-)

thr3343:
> LINES mode is still a bit fishy so don't merge it yet until we figure it out :-)

That's OK, just closed my review as it was outdated

For LINES you can try quadsIndexBuffer instead of linesIndexBuffer, but oytl tested two mods with that thus far, so haven't tested that very thoroughly

VisualEhrmanntraut:
@thr3343 using that only helps with the F3 cursor for me, MiniHUD still has some quirks

src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
VisualEhrmanntraut: Done.

xCollateral (approved):

PR #417, opened 2024-05-15
Better languages support in settings screen

ghost:
Adds missing fields for translations and uses game's ones where possible. Now more than half of the settings is translated automatically with vanilla names (Russian language for example):
image


ghost:
Didn't know that renaming a branch closes pr (duh)

Issue #181, opened 2023-01-26
Translation


forditt:
Add the lang folder to be able to add translations in the future. I wanted to translate the mod into languages such as Ukrainian and Russian, but I realized that there is no lang folder.

Felix14-v2:
Adding the lang folder will not do anything, you need to add support for text localization in the code itself for each string displayed. 
But yeah, it's cool to be able to translate the mod into our native languages.

forditt:
I understand that simply adding the lang folder will not allow adding translations, but I didn't know how to explain it correctly. Thanks for the answer!

NXTler:
I could help with translating into German.

Enderteck:
I could do the french Translation if needed when possible.

CrawKatt:
I can help with the translation into Spanish.

Versatile-One:
@forditt This should be moved to https://github.com/scully06/vulkanmod-translations

PR #421, opened 2024-05-24
Settings screen improvements

ghost:
This pr:

- Cleans up some code
- Adds all fields for translations (Uses vanilla fields, where possible, so no need to translate options, that are already translated in vanilla).
- Makes gui scale a slider
- Adds icon instead of "Video settings" text
- Moves some settings into more sensible places
- Fixes Unique opaque layer tooltip (thanks to thr)
- Fixes a bug, where max possible gui scale was unavailable
- Fixes a bug, where RangeOptionWidget played click sound when control is not hovered

Example with LOLCAT language:
image

Example of translation:
image


ghost:
Resolves #181

xCollateral (approved):

PR #426, opened 2024-05-26
Add language support for the german language

Strokkur424:
Implemented language support for the german language in VulkanMod

### DIFF
--- a/src/main/resources/assets/vulkanmod/lang/de_de.json
+++ b/src/main/resources/assets/vulkanmod/lang/de_de.json
@@ -1,0 +1,9 @@
+{
+  "vulkanmod.options.videoMode.tooltip" : "Funktioniert nur im Vollbildmodus",
+  "vulkanmod.options.windowedFullscreen" : "Fenstervollbild",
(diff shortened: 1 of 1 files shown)


NXTler:
That's a very bad translation.
"[...] Es verändert den Aspekt von einigen entfernten Blöcken" Aspekt is clearly the wrong word here, as it means something like Angle of view, direction of view, viewpoint, point of view.
And the General phrasing is too much English like, with a bad flow and oversimplifications.

NXTler:
There is also no point in doing translations without having #421 merged.

Strokkur424:
Very well. Will update the PR once #421 has been merged

Strokkur424:
> That's a very bad translation. "[...] Es verändert den Aspekt von einigen entfernten Blöcken" Aspekt is clearly the wrong word here, as it means something like Angle of view, direction of view, viewpoint, point of view. And the General phrasing is too much English like, with a bad flow and oversimplifications.

Strangely enough I agree. Eventhough I am german, the issue is, that I have no clue what the original language even ment by aspect. Once #421 has been merged I will redo some of them and figure out better translations if that is alright

NXTler:
If you want, we can collaborate once it has been merged. Just ping me here if you need another opinion or idea. I think aspect is used in the sense of "Erscheinung" in this case.

PR #432, opened 2024-06-16
[Performance Optimizations] Misc Performance and CPU Optimizations 

thr3343:
Some minor performance-related patches that should help performance in game slightly 
_(e.g. Chunk rendering is about 5-10% faster)_

* Use faster renderpass transitions
* Remove duplicated cmdBuffer constructor in CommandPool.class
* Reduce Chunk rendering CPU Overhead slightly
* Minimize native calls to waitFences via aggregated waits
* Optimize + use more precise Sampler Stage flags
* Reduce number of terrain pipelines generated
* Remove Terrain AlphaCutout uniform + hardcode to 0.5f
* Remove some unused/outdated terrain shaders/pipelines
* Some misc spec fixes


thr3343:
Decided to close this PR temporarily, as it will cause merge conflicts with a separate PR i'm working on _(Bindless rendering)_ 

That PR also has alot of the same optimizations this one has, so this PR is a bit redundant

So will likely reopen this again in the future without the merge conflicts + some new optimisations

PR #448, opened 2024-06-29
Add Simplified Chinese translation

Paper-wand:


Paper-wand:
https://github.com/scully06/vulkanmod-translations/releases

PR #465, opened 2024-07-18
[GPU Optimization/Performance] Attachment color clear skip

thr3343:
_(Based on this older PR: https://github.com/xCollateral/VulkanMod/pull/359, but with alot of the outdated/unnecessary code removed)_

A small optimization patch which automatically skips color attachment clears if the color is the same, decreasing the rate of full screen clears per frame

This helps to avoid unnecessary stalls in the GPU and improves performance slightly in game, especially at high framerates

### DIFF
--- a/src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
+++ b/src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
@@ -15,6 +15,7 @@ import org.lwjgl.opengl.GL11;
 import org.lwjgl.system.MemoryUtil;
 
 import static org.lwjgl.vulkan.VK10.*;
+import static org.lwjgl.opengl.GL11C.GL_DEPTH_BUFFER_BIT;
 
 import java.nio.ByteBuffer;
 import java.nio.FloatBuffer;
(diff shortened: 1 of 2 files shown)


src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
```diff
@@ -34,13 +35,13 @@ public abstract class VRenderSystem {
     public static int colorMask = PipelineState.ColorMask.getColorMask(true, true, true, true);
 
     public static boolean cull = true;
-
+    private static boolean canApplyClear = false;
```
Vixea: This should probably be should ;)

src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
```diff
@@ -154,12 +155,25 @@ public static MappedBuffer getShaderFogColor() {
         return shaderFogColor;
     }
 
-    public static void setClearColor(float f1, float f2, float f3, float f4) {
-        ColorUtil.setRGBA_Buffer(clearColor, f1, f2, f3, f4);
+    public static void setClearColor(float f0, float f1, float f2, float f3) {
```
Vixea: Why only change the params here just keep it the same

thr3343: The https://github.com/xCollateral/VulkanMod/pull/359 PR that this one is recycled from had some very badly formatted code, and I forgot to clean up the formatting issues as a result

Will force push the branch to correct them

src/main/java/net/vulkanmod/vulkan/VRenderSystem.java
thr3343: Renamed it from `canApplyClear` to `allowFullColorClear` instead if that's OK

Did that as `canApplyClear` is inaccurate and misleading, as only the color clear is skipped, not the depth clear

Vixea: Yea that looks good

(Noo autocorrect or me just waking up idk which one it was)

thr3343:
Replaced with https://github.com/xCollateral/VulkanMod/pull/469

PR #469, opened 2024-07-26
[GPU optimizations] Reduce attachment clears

thr3343:
_(Greatly improved version of https://github.com/xCollateral/VulkanMod/pull/465)_

A small optimization patch which removes almost all attachment clears used each frame

This helps to avoid unnecessary stalls in the GPU and improves performance slightly in game, especially at high framerates

_(If Journeymap is present, these optimizations are disabled to avoid visual glitches)_

### DIFF
--- a/src/main/java/net/vulkanmod/mixin/MixinPlugin.java
+++ b/src/main/java/net/vulkanmod/mixin/MixinPlugin.java
@@ -1,5 +1,7 @@
 package net.vulkanmod.mixin;
 
+import net.fabricmc.loader.api.FabricLoader;
+import net.vulkanmod.Initializer;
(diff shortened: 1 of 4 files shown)


thr3343:
For some reason this PR is causing performance regressions on my hardware

i.e. using LoadOpClear seems to have performance regressions over attachment clears

Which is unexpected as the spec recommends LoadOpClear over attachment clears _(as LoadOpClear allows additional driver optimizations)_

If I can't resolve this flaw, will likely close this PR to avoid/prevent introducing performance regressions into upstream

thr3343:
Decided to close in the end
Came up with some other optimizations ideas, that if successful will be much much more effective

PR #475, opened 2024-07-28
Remove GuiBatchRenderer

ghost:
Removes GuIBatchRenderer and big chunk of comment in DebugScreenOverlayM related to it, refactors ProfilerOverlay to use GuiRenderer instead, adds drawString that allow to disable shadow and fill with batching methods into GuiRenderer.


xCollateral:
While a refactor of the GUI rendering classes is needed, removing `GuiBatchRenderer` and replacing calls with `GuiRenderer` methods results in losing the batch rendering functionality.
In my opinion before removing `GuiBatchRenderer`, the batch rendering functionality should be implemented in the `GuiRenderer` class.

ghost:
> While a refactor of the GUI rendering classes is needed, removing `GuiBatchRenderer` and replacing calls with `GuiRenderer` methods results in losing the batch rendering functionality. In my opinion before removing `GuiBatchRenderer`, the batch rendering functionality should be implemented in the `GuiRenderer` class.

As you said in discord, this class was not really needed. Why keep batch rendering for GUI if all its functionality can be done with less overhead either with vanilla GuiGraphics (which uses batch rendering already AFAIK) or our GuiRenderer? I see no reason to keep these methods if they won't be used anywhere.

Also some methods are just copied from vanilla, what's the point in code duplication?

xCollateral:
Indeed it could be delete, but the batch rendering functionality might be useful. Besides the PR actually increases CPU overhead (even if it's minimal) due to the lack of batching.

PR #485, opened 2024-08-13
[Samplers] Patch: Simplify + Optimize Sampler Management

thr3343:
Simplifies sampler management by using the special hardware constant `VK_LOD_CLAMP_NONE`

`VK_LOD_CLAMP_NONE` permits the sampler to use all available miplevels on the texture, allowing one Sampler to handle all possible mipmap levels, with the additional effect of removing the need to track maxLods for each respective sampler.

Potentially may give a slight performance improvement on some hardware as a result of the reduced sampler count

### DIFF
--- a/src/main/java/net/vulkanmod/gl/GlRenderbuffer.java
+++ b/src/main/java/net/vulkanmod/gl/GlRenderbuffer.java
@@ -168,7 +168,7 @@ public class GlRenderbuffer {
             default -> 0;
         };
 
-        vulkanImage.updateTextureSampler(maxLod, samplerFlags);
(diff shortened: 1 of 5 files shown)


thr3343:
Was going to open more patch PRs like this, but released that would cause unnecessary clutter
So will close and merge this PR into a larger "Misc performance patches" PR to help tidy up + declutter the PR page

I have the code for that mostly ready to go, so it likely won't take more than a day or two.

thr3343:
Replaced by https://github.com/xCollateral/VulkanMod/pull/508

PR #491, opened 2024-08-17
[Performance] Add Early-Z optimization option (Experimental Fast Leaves)

thr3343:
Replaces the Unique Opaque Layer optimization with an experimental Early-Z option

Early-Z is a special shader optimization that culls geometry rendered behind other blocks, which has the side effect of restoring the ability to use fast leaves in VulkanMod


thr3343:
Closed to declutter PR page: Early-Z in reality had very little to no performance improvement in-game at all

_(i.e. the perfoamnce bottlenecks currently are elsewhere)_

PR #520, opened 2024-10-12
Require Fabric API 0.99.2 or above

Strokkur424:
What this PR does:
- Updates README to include the installation of Fabric API
- Modifies fabric.mod.json to add Fabric API version 0.99.2 as the lowest required version


Strokkur424:
This PR is not required, as VulkanMod instead provides FAPI library files

PR #526, opened 2024-10-14
Implement OpenGL default texture

Fox2Code:
- https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBindTexture.xhtml  
  "The value zero is reserved to represent the default texture for each texture target."

- https://stackoverflow.com/questions/15273674/binding-a-zero-texture-in-opengl  
  (Link show graphics drivers usually make it black or white, but specifics are undefined)

Also I cannot initialize the `defaultTexture` field in the static initializer because it crashes due to cycling dependencies.

This is required to properly support Not-Enough-Crashes

### DIFF
--- a/src/main/java/net/vulkanmod/gl/GlTexture.java
+++ b/src/main/java/net/vulkanmod/gl/GlTexture.java
@@ -20,6 +20,7 @@ public class GlTexture {
     private static int ID_COUNTER = 1;
     private static final Int2ReferenceOpenHashMap<GlTexture> map = new Int2ReferenceOpenHashMap<>();
     private static int boundTextureId = 0;
+    private static GlTexture defaultTexture;
     private static GlTexture boundTexture;
     private static int activeTexture = 0;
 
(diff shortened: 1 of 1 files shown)


xCollateral:
As stated by OpenGL spec:

``Texture names are unsigned integers. The value zero is reserved to represent the default texture for each texture target.``.

Your current implementation is not following that as there's only one global default texture.
At the current state it can only be used as a temporary workaround to prevent crashes in cases where the 0 value is bound.

src/main/java/net/vulkanmod/gl/GlTexture.java
```diff
@@ -35,11 +36,26 @@ public static int genTextureId() {
...
+        GlTexture defaultTexture = new GlTexture(0);
+        defaultTexture.vulkanImage = VulkanImage.createWhiteTexture();
+        GlTexture.defaultTexture = defaultTexture;
+        return defaultTexture;
+    }
+
     public static void bindTexture(int id) {
         boundTextureId = id;
-        boundTexture = map.get(id);
+        if (id == 0) {
```
xCollateral, line 52: Why not adding it directly to the map instead of doing that check?

Fox2Code: I can do that, but I would have to add a direct `if ()` statement for `glDeleteTextures`, as the documentations says for ID  deleting texture should be a noop.

Also, I cannot move image initialization to the static initializer as `VulkanImage` need the static initialize of `GlTexture` for some reason.

Fox2Code:
> As stated by OpenGL spec:
>
> `Texture names are unsigned integers. The value zero is reserved to represent the default texture for each texture target..`
>
> Your current implementation is not following that as there's only one global default texture.
> At the current state it can only be used as a temporary workaround to prevent crashes in cases where the 0 value is bound.

Well, I implemented as close to my understanding as possible, if I'm that far from the OpenGL spec, this means I for sure don't understand what the OpenGL spec want to say by using thoses words.

I used this exact statement to guide the implementation included in this PR.

src/main/java/net/vulkanmod/gl/GlTexture.java
Fox2Code: OpenGL Documentation for `glDeleteTextures`:
- https://registry.khronos.org/OpenGL-Refpages/gl4/html/glDeleteTextures.xhtml

Fox2Code:
Note: @xCollateral want to do the commit himself, I keep this PR as a reminder for him.

Versatile-One:
> Note: @xCollateral want to do the commit himself, I keep this PR as a reminder for him.

Has collateral done the commit? I would have asked collateral but it seems like he is very busy atm.

PR #535, opened 2024-10-27
Add Korean Translation

relpia:


NXTler:
Translations will probably not be merged.

trnahuman:
Vkmod prefers not to accept translations, so I suggest you post your translation here, which is developed by the community https://github.com/scully06/vulkanmod-translations

PR #546, opened 2024-11-23
Update Vulkan.java

ItsWinterBush26:


NXTler:
That's not how you downgrade to VulkanAPI 1.1, or make it work with Pojav if that is your plan.

PR #614, opened 2025-05-14
Fix F3 + L (default profiler) crash

Haberno:
When the F3+L is pressed it triggers vanilla profiler, it tries to use these OpenGL-specific TimerQuery objects which crashes the game,


1mshy:
y

PR #617, opened 2025-05-25
Fix Entity Distance 0 value when below 100

Ryvk08:
This will fix a minor bug with Entity Distance options with VulkanMod!


Kichura (approved):

Versatile-One:
@Ryvk08 This should be closed as this issue has been fixed in vulkanmod some times ago

PR #631, opened 2025-06-16
add pt-br localization

brviictoor:
this mod is awesome and to make it easier for brazilians to use, added a pt-br translation.


Versatile-One:
@brviictoor This should be moved to https://github.com/scully06/vulkanmod-translations

PR #671, opened 2025-09-12
Update fabric.mod.json

mine1x:


NXTler (commented):
1.21.8 is not the current version of VulkanMod, this will not work.

PR #696, opened 2025-12-13
Added a system property to skip wayland compatibility patches

amiralimollaei:
Added system property "vkmod.wayland.compat.skip" that when set, disables patches by VulkanMod that are made to prevent errors and crashes when running under Wayland, the reason for this is to enable advanced users to provide their own patched versions of `libglfw.so` that adds native support for Wayland, fixing these crashes outside of the context of this mod.

Wayland users can set `-Dorg.lwjgl.glfw.libname=/path/to/custom/glfw/libglfw.so` and `-Dvkmod.wayland.compat.skip` to enable full Wayland compatibility, provided they use [this patched version of libglfw](https://github.com/BoyOrigin/glfw-wayland), fixing many bugs when running under Wayland.

### DIFF
--- a/src/main/java/net/vulkanmod/config/Platform.java
+++ b/src/main/java/net/vulkanmod/config/Platform.java
@@ -10,6 +10,7 @@ import static org.lwjgl.glfw.GLFW.*;
 public abstract class Platform {
     private static final int activePlat = getSupportedPlat();
     private static final String activeDE = determineDE();
+    public static final boolean skipWaylandPatches = shouldSkipWaylandPatches();
 
     public static void init() {
         // Increase stack size to 256 KB to prevent out of stack error on nvidia driver
@@ -63,6 +64,11 @@ public abstract class Platform {
             return xdgCurrentDesktop.toLowerCase();
         return "N/A";
     }
+
+    private static boolean shouldSkipWaylandPatches() {
+        return System.getProperties().containsKey("vkmod.wayland.compat.skip");
+    }
+
 
     public static int getActivePlat() {
         return activePlat;
--- a/src/main/java/net/vulkanmod/config/Platform.java
+++ b/src/main/java/net/vulkanmod/config/Platform.java
@@ -10,7 +10,6 @@ import static org.lwjgl.glfw.GLFW.*;
 public abstract class Platform {
     private static final int activePlat = getSupportedPlat();
     private static final String activeDE = determineDE();
-    public static final boolean skipWaylandPatches = shouldSkipWaylandPatches();
 
     public static void init() {
         // Increase stack size to 256 KB to prevent out of stack error on nvidia driver
@@ -65,10 +64,9 @@ public abstract class Platform {
         return "N/A";
     }
 
-    private static boolean shouldSkipWaylandPatches() {
+    public static boolean shouldSkipWaylandPatches() {
         return System.getProperties().containsKey("vkmod.wayland.compat.skip");
     }
-
 
     public static int getActivePlat() {
         return activePlat;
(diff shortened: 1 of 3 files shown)


src/main/java/net/vulkanmod/config/Platform.java
```diff
@@ -64,6 +65,11 @@ private static String determineDE() {
         return "N/A";
     }
 
+    private static boolean shouldSkipWaylandPatches() {
+        return System.getProperties().containsKey("vkmod.wayland.compat.skip");
+    }
+
+
```
Strokkur424: Remove the extra space here just for style reasons.

src/main/java/net/vulkanmod/config/Platform.java
```diff
@@ -10,6 +10,7 @@
 public abstract class Platform {
     private static final int activePlat = getSupportedPlat();
     private static final String activeDE = determineDE();
+    public static final boolean skipWaylandPatches = shouldSkipWaylandPatches();
```
Strokkur424: Generally not a huge fan of the static fields here. Maybe make the method itself public and remove this field. This has the advantage that it is override-able and just generally nicer to work with.

Strokkur424 (changes requested):

amiralimollaei:
okay, I've made method itself public and now it directly calls the method.

src/main/java/net/vulkanmod/config/Platform.java
```diff
@@ -64,6 +64,10 @@ private static String determineDE() {
         return "N/A";
     }
 
+    public static boolean shouldSkipWaylandPatches() {
```
Strokkur424, line 67: I am not really happy with the method name tbh. "Should skip" typically means to *not* do something, but in the use cases for this method you actually "don't skip" stuff. Maybe rename the method to `hasPatchedWayland`?

Strokkur424 (changes requested):

src/main/java/net/vulkanmod/config/Platform.java
amiralimollaei: No, because then it would mean we ARE patching, when in reality the name means we are skipping the patches, though the patches just happen to skip lines of code, that means although we are doing extra when we skip the patches, we literally do exactly as if the patch didn't exist, therefore we do nothing (or nothing changes), meaning we skip the patches. Which is the name of the method.

Strokkur424: I feel like it still reads nicer in source code. Regardless of whether the skipping is the patching, which are skipping (skipping the skip); I think it just makes it clearer.

amiralimollaei: I feel like that reads wrong, like I would imagine a different functionality where it does the exact opposite of what it does now, if I changed the name

NotNekodev: i would say that it would sound better, when it was something like "shouldPatchWayland()", but that would also mean that you had to invert all uses so like where you have false you would need to make it true and vice versa to fit the naming

amiralimollaei: shouldPatchWayland kinda implies that you also have to know if you're on wayland or not, maybe i could just move the entire condition of applying patches to the function?
like if platform is wayland and if the vkmod.wayland.compat.skip doesn't exist

Versatile-One:
@amiralimollaei Is this still relevant now?

amiralimollaei:
No, the custom glfw libraries don't work anymore on the latest version.

Issue #198, opened 2023-03-06
Colors are slightly muted on macOS


cadenkriese:
On macOS, the games colors appear slightly muted when compared to the vanilla game. All screenshots contain an instance running Sodium on the left and VulkanMod on the right. Sodium behaves the same as vanilla, I didn't realize it was modded when I took the screenshots

## Example 1
<img src="https://user-images.githubusercontent.com/13630061/223187461-1650e4a5-2675-4628-bc3b-d407c50c6687.png" width="550" alt="Highlighted RGB color shows (255,188,0)" />

*Sodium is highlighted in Digital Color Meter.*

<img src="https://user-images.githubusercontent.com/13630061/223187948-d114a939-bc6f-4eeb-95e6-a2af4ff19ce5.png" width="550" alt="Highlighted RGB color shows (251,191,36)" />

*The same spot in VulkanMod is highlighted in Digital Color Meter.*

## Example 2

<img src="https://user-images.githubusercontent.com/13630061/223188192-1d7a5fba-0c66-426d-ba4c-83373817cba1.png" width="550" alt="Highlighted RGB color shows (123,80,31)" />

*Sodium is highlighted in Digital Color Meter.*

<img src="https://user-images.githubusercontent.com/13630061/223188291-d51d59a7-7d19-4d8c-845e-82c21c052e00.png" width="550" alt="Highlighted RGB color shows (108,76,52)" />

*Approximately same spot in VulkanMod is highlighted in Digital Color Meter.*

Splarkszter:
are you using an M1(or of the same family) processor?

i remember people have had graphics issues with those on Java applications.

cadenkriese:
> are you using an M1(or of the same family) processor?
> 
> 
> 
> i remember people have had graphics issues with those on Java applications.

On an M1 Max, but the issue is with VulkanMod since it produces different results than vanilla on the same hardware.

Jubster7:
how are you getting the game to run on Mac? I tried to get it to work but keep getting the error code : 

The game crashed whilst initializing game
Error: java.lang.UnsatisfiedLinkError: Failed to locate library: libvulkan.1.dylib

(I already installed the Vulcan SDK with moltenvk)

edit: figured it out , have to download the Mac specific version on modrinth : https://modrinth.com/mod/vulkanmod/version/BXiDYSte

Git-North:
@Jubster7 wow im so dumb had the same problem now i figured it out

mintexists:
[...]

PR #719, opened 2026-01-24
Reproduce color behavior on macOS P3 displays

TNTpower0v0:
Hi! I did this with the help of AI, but I wrote the code myself. Everything is working well on MacOS, and my friend with Windows and Linux has also confirmed normal colors.
- Adds native Display P3 color space support for modern mac displays (fallback for compatibility for older/differen conf. macs) 
- Enables` GLFW_COCOA_RETINA_FRAMEBUFFER` to guarantee native resolution rendering on Retina displays
- Extended Color Depth. 10-bit color (10+10+10+2 format)

### DIFF
--- a/src/main/java/net/vulkanmod/mixin/window/WindowMixin.java
+++ b/src/main/java/net/vulkanmod/mixin/window/WindowMixin.java
@@ -61,6 +61,15 @@ public abstract class WindowMixin {
     private void vulkanHint(WindowEventHandler windowEventHandler, ScreenManager screenManager, DisplayData displayData, String string, String string2, CallbackInfo ci) {
         GLFW.glfwWindowHint(GLFW_CLIENT_API, GLFW_NO_API);
 
+        // Enable 32bit P3 wide color gamut on macOS
+        if (Platform.isMacOS()) {
+            GLFW.glfwWindowHint(GLFW_COCOA_RETINA_FRAMEBUFFER, GLFW_TRUE);
+            GLFW.glfwWindowHint(GLFW_RED_BITS, 10);
+            GLFW.glfwWindowHint(GLFW_GREEN_BITS, 10);
+            GLFW.glfwWindowHint(GLFW_BLUE_BITS, 10);
+            GLFW.glfwWindowHint(GLFW_ALPHA_BITS, 2);
+        }
+
         //Fix Gnome Client-Side Decorators
         boolean b = (Platform.isGnome() | Platform.isWeston() | Platform.isGeneric()) && Platform.isWayLand();
         GLFW.glfwWindowHint(GLFW_DECORATED, (b ? GLFW_FALSE : GLFW_TRUE));
--- a/src/main/java/net/vulkanmod/vulkan/Vulkan.java
+++ b/src/main/java/net/vulkanmod/vulkan/Vulkan.java
@@ -110,19 +110,11 @@ public class Vulkan {
 
     }
 
-    public static VkDevice getVkDevice() {
-        return DeviceManager.vkDevice;
-    }
-
-    public static long getAllocator() {
-        return allocator;
-    }
-
     public static long window;
-
     private static VkInstance instance;
     private static long debugMessenger;
     private static long surface;
+    private static List<String> supportedInstanceExtensions;
 
     private static long commandPool;
     private static VkCommandBuffer immediateCmdBuffer;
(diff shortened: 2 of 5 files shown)


cadenkriese:
Fixes #198 for me

tankonatorinweb:
It’s working for me

tankonatorinweb:
@xCollateral is it possible?

TNTpower0v0:
@BloCamLimb so its absolute incorrect to make?

BloCamLimb:
I don't think this PR actually fixes the root cause; in fact, it's essentially replicating a known color management flaw.

The "vibrancy" experienced in the OpenGL version (Vanilla/Sodium) is a side effect of a long-standing GLFW bug https://github.com/GLFW/glfw/issues/2748 (or rather, a lack of implementation) where it fails to call `setColorSpace`. On macOS, this causes the WindowServer to treat raw sRGB values as if they were already in the **Display P3** gamut. This "stretches" the colors, resulting in over-saturation that might look "better" to some, but is technically a bug.

The "muted" look in Vulkan is actually the **correct, calibrated behavior**. The system is properly remapping sRGB values to the P3 display so that a red in your app looks like the *intended* red, not an over-saturated version of it.

By simply switching the Swapchain to P3 without adding a proper color space conversion pass in the shader, this PR is just **matching one mistake with another**. You're interpreting sRGB data as P3 coordinates, which is mathematically incorrect and breaks platform consistency.

**The implications of merging this:**

* Loss of Accuracy: We lose the ability to display "true" sRGB colors.
* Platform Inconsistency: The app will look completely different on a P3 Mac compared to a standard sRGB monitor or Windows/Linux, where the sRGB-to-P3 "stretching" won't happen.
* The "Fake" Fix: We are choosing a "pleasing" bug over a correct implementation.

If the goal is to truly support the P3 Wide Gamut, we shouldn't just toggle the Swapchain. A real wide-gamut implementation requires a much deeper architectural shift:

* P3-Aware Offscreen Buffers: All offscreen render targets and intermediate buffers should be set to the P3 color space to prevent clipping.
* High Precision Textures: We should utilize RGBA16Float (F16) textures to provide the necessary headroom and precision for the expanded color volume, avoiding banding in the wider gamut.
* Proper Color Transform: We must implement a dedicated color management pass in the fragment shader to transform coordinates from sRGB to P3.

Without these steps, we aren't "supporting" P3; we're just abusing the wide gamut to mask an sRGB calibration issue. I suggest we revert this (since this is already cherry-picked) and stick to a standard sRGB workflow until we are ready to implement a full-pipeline wide-gamut solution.


> @BloCamLimb so its absolute incorrect to make?

Yes

TNTpower0v0:
@BloCamLimb Thx for comment

TNTpower0v0:
@BloCamLimb But i still think some people playing vanilla with broken colors and just used to it. For the same reason, I didn't notice the incorrect colors and ended up repeating the mistake.

BloCamLimb:
@TNTpower0v0 This issue should be reported to Mojang. If they don't address this legacy behavior, players will continue to mistake broken (stretched) colors for "correct" ones.

This is quite easy to fix in OpenGL side, adding these few lines of code during window initialization
```obj-c
NSWindow *window = (NSWindow *) glfwGetCocoaWindow(glfwWindow);
[window setColorSpace:[NSColorSpace sRGBColorSpace]];
```
```java
long windowPtr = GLFWNativeCocoa.glfwGetCocoaWindow(glfwWindow);
NSObject window = new NSObject(new Pointer(windowPtr));
Client client = Client.getInstance();
Object srgbSpace = client.sendProxy("NSColorSpace", "class").sendProxy("sRGBColorSpace");
window.send("setColorSpace:", srgbSpace);
```

cadenkriese:
Thanks for finding the root cause that's very interesting. I'm in favor of merging this PR anyway to match vanilla behavior because that's the way the game has looked on Mac for years, and the one users have come to expect. Plus if the glfw issue gets fixed in the future then VulkanMod can easily revert and match. 

Full P3 support is a different issue that I personally don't care about until after VulkanMod support custom shaders (I believe Iris has this).

xCollateral:
I agree with @BloCamLimb, the game is meant to be rendered in sRBG color space, mapping the output as is to P3 isn't a correct solution.

Anyway this PR was meant to reproduce vanilla behavior, which macOS users have come to expect. I don't think that forcing P3 is correct, the only workaround for now might be to let users choose which color space they want to use.

TNTpower0v0:
Closed due mojang fixed colors in 1.26.2 for both OpenGL and Vulkan api (correct sRGB) there now no need in PR.

PR #759, opened 2026-03-16
Use correct rendering device name for Blaze3D

whitebelyash:
Replace VulkanMod version string with the rendering device name. This brings used GPU name in F3/Debug HUD back.

I've moved the VulkanMod string down to the API version. Still not sure if it's a correct choice, let me know if it's not!

Looks like this now:
image


whitebelyash:
Fixed in upstream.

PR #788, opened 2026-04-12
Add flake.nix for devshell

indium114:
As it says on the tin

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ bin/
 # fabric
 
 run/
+
+# direnv
+/.direnv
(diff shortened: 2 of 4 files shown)


NXTler:
I'm not familiar with NixOS and don't really see the purpose for this. Could you please elaborate?

indium114:
> I'm not familiar with NixOS and don't really see the purpose for this. Could you please elaborate?

A devshell is essentially a temporary environment which has all of the tools needed for development, which is needed because NixOS discourages polluting your normal environment with development tools. This PR adds a `flake.nix` file, which outputs a devshell with all of the development tools needed to work on this project.

The `.envrc` file works with Direnv, so that you don't have to type `nix develop` to enter the devshell, it will automatically enter it when you `cd` into the project directory.

The `.gitignore` entry ignores the `.direnv` directory, which is a cache for the devshell when used with Direnv

PR #797, opened 2026-04-20
Fix Ctrl+key combos inserting characters in text fields on Wayland

Tmpod:
On native Wayland, GLFW fires a separate char event after the "main" key event, leading to quirky behaviour (eg. with `Ctrl+a`). This new mixin fixes it by cancelling the secondary key event if the Ctrl modifier is set.

_I'm doing the PR against `dev`, which still seems to be on 1.21.10. What's up with that?_


NotNekodev:
lgtm, tested and works perfectly

xCollateral (approved):

Tmpod:
Hey thanks! Do I need to do anything else here?

Issue #791, opened 2026-04-14
Fix this fucking crash pls!


AutonomiusUltraInstinct:
[Bug] VK_ERROR_DEVICE_LOST when switching to Fullscreen (F11) in PvP

Pls fix this shit [it also saying this in event logger]:
Unable to find a description for event ID 153 from source nvlddmkm. The component that raises this event is not installed on this local computer or is corrupted. Install or repair the component on the local computer.

If the event occurred on a different computer, you may need to save the display details along with the event.

The following details were added to the event:

\Device\000000b2
Error occurred on GPUID: 100

The message resource exists, but the message was not found in the message table.
every fucking time when i am in pvp and i am doing f11 it can crash

Confiac1563:
Can you try updating the GPU driver?

KoyuKansou:
how about you upload the latest.log instead and try updating the gpu driver and don't be a jerk

Issue #724, opened 2026-02-04
[1.21.11 - VulkanMod 0.5.10] Game Crash on Start with VanillaTweaks Resource Packs


The-Letter-M:
### Describe the bug

Minecraft crashes on start up when any Vanilla Tweaks resource pack is loaded and already enabled, specifically when Remove Reloading Screens is enabled. Without RRS, the game starts and instead freezes upon attempting to load the resource packs. Attempting to reload Resource Packs from this point with F3 + T leads to a similar crash. After the crash, all resource packs are disabled and must be re-enabled.

Minecraft is able to run when the pack is disabled before launch and then re-enabled after launch. Other Resource Packs (such as Bare Bones or Redstone Tweaks) seem to function perfectly fine.

Tested on Hyprland Arch Linux, Nvidia RTX 3060.

### Steps to reproduce

1. Have any Vanilla Tweaks Resource Pack loaded and enabled.
2. Enable VulkanMod.
3. Attempt to launch the game.
4. Game will either crash on startup (if Remove Reloading Screens is enabled) or fail to load resource packs.

### Log files

When Remove Reloading Screens is enabled:
[crash-2026-02-04_16.32.39-client.txt](https://github.com/user-attachments/files/25082441/crash-2026-02-04_16.32.39-client.txt)

With RRS disabled and refreshing textures after the screen freezes:
[crash-2026-02-04_16.45.08-client.txt](https://github.com/user-attachments/files/25082463/crash-2026-02-04_16.45.08-client.txt)

mjothr:
it seems to be an issue with vanilla tweaks' pixel consistency features, or different texture resolutions as a whole according to some peoples errors. i removed the pixel consistency features and it launched for me

The-Letter-M:
Hm, I disabled every module in the Fixes and Consistency section on Vanilla Tweaks and I still cannot launch the game with the resource pack already enabled. I can't seem to figure out which other modules are causing the crashing, but it's more than just the consistency fixes causing the issue.

Versatile-V1:
@The-Letter-M Can you send the vanilla tweaks resource pack you are using?

The-Letter-M:
@JJ10-GNU Of course, sorry for the late reply.
Here ya go.

[VanillaTweaks_lite_MC1.21.x.zip](https://github.com/user-attachments/files/25312918/VanillaTweaks_lite_MC1.21.x.zip)

Versatile-V1:
[...]

Issue #708, opened 2026-01-08
Mod Menu Incompatible on Linux


CaptainDeSync:
### Describe the bug

Mod Menu causes the game to crash on launch on Linux specifically. Windows its fine somehow.

### Steps to reproduce

Add Vulkan and Mod Menu and try to play together. The game should crash.

### Log files

https://mclo.gs/fySU9g0

Versatile-V1:
Are you on Wayland?

CaptainDeSync:
> Are you on Wayland?

Yes!

Versatile-V1:
That's the issue I think. Because I am on x11 and it works fine

CaptainDeSync:
> That's the issue I think. Because I am on x11 and it works fine

How weird. Good to know!

GrimPlayzs:
Also on Wayland same issue.

kxxzo:
Same on windows (11) btw

FerryIsCaring67:
Same for windows 11

MotWnb:
Same for windows 11

Versatile-V1:
> ### Describe the bug
> 
> Mod Menu causes the game to crash on launch on Linux specifically. Windows its fine somehow.
> ### Steps to reproduce
> 
> Add Vulkan and Mod Menu and try to play together. The game should crash.
> ### Log files
> 
> https://mclo.gs/fySU9g0

@CaptainDeSync The log you sent says other things though

Ok so basically its a out of stack space error (only happens with mod menu it looks like). To fix this add -Dorg.lwjgl.system.stackSize=256
to your jvm arguments.

retcinder:
I don't experience this on Wayland (using 0.5.8), perhaps another thing is causing it?

GrimPlayzs:
If anyone is having issues with Mod Menu with VulkanMod try adding ```-Dorg.lwjgl.system.stackSize=256``` to your JVM arugments

heav33n:
it appears that the mod is broken when running under wayland

GrimPlayzs:
how so?

retcinder:
> it appears that the mod is broken when running under wayland

Untrue, runs perfectly fine on my machine,

Versatile-V1:
This issue should be fixed now

PR #798, opened 2026-04-20
Fix incorrect handling of VkResult in swapchain creation

DissolveDZ:
Hi,
I ran into some problems running this mod on my wayland linux setup and after some investigation it turned out the error handling of VkResult is incorrect.
Basically in Vulkan.java the checkResult function:
```java
public static void checkResult(int result, String errorMessage) {
  if (result != VK_SUCCESS) {
    throw new RuntimeException(String.format("%s: %s", errorMessage, VkResult.decode(result)));
  }
}
```
should be changed to 
```java
public static void checkResult(int result, String errorMessage) {
  if (result < VK_SUCCESS) {  // Change to <
    throw new RuntimeException(String.format("%s: %s", errorMessage, VkResult.decode(result)));
   }
}
```
Fixes:
https://github.com/xCollateral/VulkanMod/issues/791
https://github.com/xCollateral/VulkanMod/issues/724
https://github.com/xCollateral/VulkanMod/issues/708

This prevents fatal crashes on many linux setups. VkResult codes like VK_SUBOPTIMAL_KHR or VK_TIMEOUT should NOT be treated as fatal according to the vulkan spec.
Fatal errors return negative values, while non-fatal errors return positive values.
Ref: 
https://docs.vulkan.org/refpages/latest/refpages/source/vkCreateSwapchainKHR.html
https://docs.vulkan.org/refpages/latest/refpages/source/VkResult.html

Tested on multiple setups, this is the correct behavior and "fixes" non-fatal errors being treated as crashes.
Errors like temporary out of date errors will resolve without crashing (e.g. resizing a window, losing window focus, etc.).

### DIFF
--- a/src/main/java/net/vulkanmod/vulkan/Vulkan.java
+++ b/src/main/java/net/vulkanmod/vulkan/Vulkan.java
@@ -381,7 +381,7 @@ public class Vulkan {
     }
 
     public static void checkResult(int result, String errorMessage) {
-        if (result != VK_SUCCESS) {
+        if (result < VK_SUCCESS) {
             throw new RuntimeException(String.format("%s: %s", errorMessage, VkResult.decode(result)));
         }
     }


NotNekodev:
lgtm

R00tB33rMan (approved):

PR #822, opened 2026-05-19
Persistent Staging Buffer Upload

futz12:
- [x] Persistent Staging Buffer Upload


futz12:
Persistent Staging Buffer Upload:

Render distance: 32Chunk

Before(6.6): 
image

After:
image

Versatile-One:
@futz12 Wouldn't it be better to have separate pull requests for each optimization? Or are they interconnected?

futz12:
> @futz12 Wouldn't it be better to have separate pull requests for each optimization? Or are they interconnected?

ok

PR #835, opened 2026-06-04
Update build.gradle

Syaltx:
Linux Arm64 Support.


NXTler:
They are alread in "VulkanMod Additional Libraries" iirc, there is no need to put them in VulkanMod directly.
This can be closed.

PR #839, opened 2026-06-07
[Fix] Issues in Video Option Screen

amiralimollaei:
Fixes two bugs in the video option screen:

1. Clicking on the mod name/icon has a clicking sound but does nothing
2. if the mod's name is too long, it extends outside of the area assigned to the `ModIconWidget`

for the second issue, I've made the text pan to left and right (animation based on system time) and used scissors so the text wont exceed the boundaries assigned to  `ModIconWidget`

Before:
image

After:
Screenshot From 2026-06-07 22-34-56


amiralimollaei:
I've updated the PR to use the built in text-panning functionality similar to `CyclingOption`

xCollateral (approved):

PR #850, opened 2026-06-22
[feature] Implement scrollable page list

amiralimollaei:
This PR refactors the options screen GUI by extracting scrolling code from `VOptionList` into a new abstract base class `AbstractScrollList`, and replacing the page button layout in `VOptionScreen` with a dedicated, scrollable `VPageList` gui element.

> [!WARNING]
> I used Deepseek for parts of this PR, seems easy enough that even AI can do it, that doesn't mean it's wrong, I've verified the code manually.

## Checklist

- [x] This PR doesn't break already existing mods that register to the Video Options GUI
- [x] Verified that the options list scrolling feels identical to pre-refactor behavior.
- [x] Verified that the new page sidebar scrolls correctly when populated with a high volume of mod entries.
- [x] Ensured scissor clipping bounds are accurate and don't bleed into adjacent UI elements.

### DIFF
--- a/src/main/java/net/vulkanmod/config/gui/AbstractScrollList.java
+++ b/src/main/java/net/vulkanmod/config/gui/AbstractScrollList.java
@@ -1,0 +1,297 @@
+package net.vulkanmod.config.gui;
+
+import com.mojang.blaze3d.opengl.GlStateManager;
+import it.unimi.dsi.fastutil.objects.ObjectArrayList;
+import net.minecraft.client.gui.components.events.GuiEventListener;
+import net.minecraft.client.input.MouseButtonEvent;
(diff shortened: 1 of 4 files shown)


src/main/java/net/vulkanmod/config/gui/AbstractScrollList.java
```diff
@@ -0,0 +1,297 @@
...
+            return false;
+        }
+
+        @Override
+        public void setFocused(boolean bl) {
+            if (widget != null)
+                widget.setFocused(bl);
+        }
+    }
+}
```
amiralimollaei, line 297: Moved from VOptionList, to be shared with VPagesList

PR #860, opened 2026-07-02
chore: ignore agent runtime artifacts

enchan087:


NXTler:
There is zero need to have this in the main repository as we do not use AI, nor do we want to encourage its use.

Issue #674, opened 2025-09-22
Xaero's World Map and Xaero's Minimap are not compatible. Add to wiki?


DarthJahus:
Doesn't prevent start, but crashes when entering a world.

World Map 1.39.12:
```
Caused by: java.lang.IllegalStateException: No GLCapabilities instance set for the current thread.
```

Mini Map 25.2.10:
```
Caused by: java.lang.IllegalStateException: Unsupported non-OpenGL rendering detected!
at xaero.common.graphics.renderer.multitexture.MultiTextureRenderTypeRendererProvider.defaultTextureBind
```

Tested on 1.21.5. Fabric 0.17.2.

FASTB1mc:
It is in the wiki. https://github.com/xCollateral/VulkanMod/wiki/Incompatible-mods

DarthJahus:
Weird. I searched the page with CTRL+F. Nevermind.

PR #870, opened 2026-07-15
Fix crash on renderbuffers with no sampler filter (#674)

Chaos02:
`VkGlRenderbuffer` treats a GL renderbuffer like a sampled texture and demands a min filter. A renderbuffer is a render target with no sampler, so `minFilter` stays at its unset default of 0 and `updateSampler()` falls through to `throw "Unexpected min filter value: 0"`. Treat `minFilter == 0` as "no sampler needed" and return early.

Triggered in the wild by Xaero's `ImprovedFramebuffer` (issue #674). Confirmed fixed in-game on MC 1.21.1 with Xaero's Minimap / World Map.

Scope note: I originally paired this with a framebuffer deferred-attach guard, but that only turns the crash into wrong-texture map rendering — Xaero's map needs a proper deferred attachment-resolve in `VkGlFramebuffer`, which I'm working on separately. This PR is just the standalone renderbuffer crash fix.

### DIFF
--- a/src/main/java/net/vulkanmod/gl/VkGlRenderbuffer.java
+++ b/src/main/java/net/vulkanmod/gl/VkGlRenderbuffer.java
@@ -156,6 +156,14 @@ public class VkGlRenderbuffer {
     }
 
     void updateSampler() {
+        // A renderbuffer is a render target, not a sampled resource, so it may
+        // legitimately have no sampler filter set (minFilter == 0, the unset
+        // default). Treat that as "no sampler needed" and skip the sampler
+        // build instead of falling through to the "Unexpected min filter value"
+        // throw below. See issue #674 (Xaero's custom depth renderbuffer FBO).
+        if (minFilter == 0)
+            return;
+
         if (vulkanImage == null)
             return;
 


Chaos02:
Scope note while this is open, since it is Xaero motivated:

The renderbuffer crash fixed here (#674), and the FBO deferred-attach issue mentioned in the description, are both legitimately VulkanMod's side. VulkanMod emulates GL framebuffers, so it should tolerate an FBO built the way Xaero builds one (attach-before-allocate).

There is a third Xaero-on-VulkanMod crash that I would call out of scope for VulkanMod though. Xaero's World Map calls `org.lwjgl.opengl.GL.getCapabilities()` from `MapLimiter.determineDriverType` (driver / VRAM detection) and dies with `IllegalStateException: No GLCapabilities instance set for the current thread` under VulkanMod. A Vulkan backend has no GL context to report capabilities for, so that one belongs on Xaero's side rather than being emulated here.

Xaero's trackers, for cross reference:

- Xaero's World Map: https://www.curseforge.com/minecraft/mc-mods/xaeros-world-map (issues: https://www.curseforge.com/minecraft/mc-mods/xaeros-world-map/issues)
- Xaero's Minimap: https://www.curseforge.com/minecraft/mc-mods/xaeros-minimap (issues: https://www.curseforge.com/minecraft/mc-mods/xaeros-minimap/issues)

Flagging it so it does not get lumped into VulkanMod's Xaero compat surface.

Chaos02:
Related PRs from the same effort (getting a VulkanMod based 1.21.1 pack working):

- #871 (independent of this one)
- #872 (same Xaero investigation as this PR)

One correction to the scope note in the description above. I said Xaero's map needed a proper deferred attachment resolve in `VkGlFramebuffer`, and that I was working on it separately. That turned out to be wrong. I built that fix and traced it in game: the FBO lifecycle is actually healthy, and the map was still corrupted. The real cause was `UNPACK_ROW_LENGTH == 0` being taken literally, which is what #872 fixes.

The deferred attach case is still real (a texture attached to an FBO before its storage is allocated does throw "Image is null"), it is just a separate crash rather than the map corruption. Happy to put that up separately if it is wanted.

PR #871, opened 2026-07-16
Fix crash when a graphics pipeline is bound with no active render pass

Chaos02:
Fixes an intermittent `NullPointerException` that kills the game when the window is
occluded, minimized, or the title bar is being dragged.

## Symptom

```
java.lang.NullPointerException: Cannot invoke "...RenderPass.getId()" because "state.renderPass" is null
	at net.vulkanmod.vulkan.shader.GraphicsPipeline.createGraphicsPipeline(GraphicsPipeline.java:192)
	at ...Object2LongOpenHashMap.computeIfAbsent (getHandle:58)
	at net.vulkanmod.vulkan.Renderer.bindGraphicsPipeline(Renderer.java:638)
	at net.minecraft.class_5944.bindPipeline(ShaderInstance)
```

## Repro

In-game, open chat, then tab out or minimize the window. Crashes intermittently.

It is a race between three conditions: the window stops having a valid render target,
the GUI (chat) keeps rendering anyway, and a GUI shader is used whose pipeline is not
yet in the cache.

## Cause

The GUI still renders while no render pass is bound. `bindGraphicsPipeline` builds a
`PipelineState` from the current (null) `boundRenderPass`; the first use of an
uncached GUI shader then reaches `createGraphicsPipeline`, which dereferences
`state.renderPass.getId()` and throws. A cached pipeline survives this path, which is
why the crash only shows up when a not-yet-compiled shader happens to be used during
the occluded frames.

## Fix

Skip the bind when no render pass is bound. Without a pass there is no valid target to
record a draw into, so there is nothing useful to do in this method anyway.

This covers the whole path to the NPE:

- `createGraphicsPipeline` has two call sites.
- Site 2, `getHandle`, is only reached from `Renderer.bindGraphicsPipeline`, which this
  guard covers.
- Site 1, the eager compile in the `GraphicsPipeline` constructor, is gated on
  `Pipeline.Builder.renderPass`.

## Notes

One unrelated observation found while tracing this, left alone here: `PipelineState.DEFAULT`
carries a null `renderPass`, so if the constructor's eager-compile path were ever wired up it
would hit the same dereference on non-`DYNAMIC_RENDERING` devices. Not reachable today; happy
to open a separate issue if that is worth tracking.

## Verification

Repro steps above against the patched build: no crash across repeated tab-outs.

### DIFF
--- a/src/main/java/net/vulkanmod/vulkan/Renderer.java
+++ b/src/main/java/net/vulkanmod/vulkan/Renderer.java
@@ -632,6 +632,14 @@ public class Renderer {
     }
 
     public void bindGraphicsPipeline(GraphicsPipeline pipeline) {
+        // No active render pass, e.g. the GUI still renders while the window is
+        // occluded or the title bar is being dragged. Binding here would build a
+        // PipelineState with a null render pass and NPE in
+        // GraphicsPipeline.createGraphicsPipeline at state.renderPass.getId().
+        // There is nothing to record into without a pass, so skip the bind.
+        if (boundRenderPass == null)
+            return;
+
         VkCommandBuffer commandBuffer = currentCmdBuffer;
 
         PipelineState currentState = PipelineState.getCurrentPipelineState(boundRenderPass);


Chaos02:
Related PRs from the same effort (getting a VulkanMod based 1.21.1 pack working). Both are independent of this one, which is purely the tab-out path:

- #870
- #872

Chaos02:
I am withdrawing the flash half of this PR. It is wrong, and the verification I claimed for it
does not hold up. Detail below, since the reasoning matters more than the retraction.

**The stated rationale is factually wrong.** The PR body claims `LOAD_OP_LOAD` on the main
swapchain pass "retains the previously presented frame". It does not. With N swapchain images,
`LOAD` on a swapchain image yields what *that image* held N frames ago, not what was presented
last frame. In game this is immediately visible once the change is actually live: the panorama,
the resource-pack MOTD splash and the mod-update indicator all jitter back and forth through N
stale states while only slowly advancing. It is every animated GUI element, not just the
background I was aiming at.

**It also reads undefined memory after a swapchain recreate.** `SwapChain.createSwapChain`
builds fresh `VulkanImage`s on each recreate, so their contents are UNDEFINED on first use,
and `LOAD` then loads undefined. Window drag and the GNOME overview both trigger recreation.
This would explain the black flashes I now see once the change is live, and the cadence fits
(first tap clean, drag brings on black and jitter, every subsequent tap has it), though I
have not yet finished isolating those artifacts against a stock build.

**The "Verified in-game on MC 1.21.1" claim is not supportable, and I should not have made
it.** What I actually tested was a backport of this change in a companion mod on 1.21.1, not
this patch against `dev`, and on 1.21.1 neither half of the flash fix ever executed:

- A runtime trace of every `Renderer.clearAttachments` call on 1.21.1 shows **zero**
  `DEPTH|COLOR` (`0x4100`) clears. The `preFrameOps` frame-start `DEPTH|COLOR` clear that the
  clear-strip targets does not exist in the shipped 1.21.1 version, so the strip never
  matched anything. The only COLOR clear on 1.21.1 is a COLOR-only `0x4000` from
  `LoadingOverlay.render`, which the strip deliberately passes through.
- The `DefaultMainPass` `LOAD` was shadowed: the test profile had Beryl installed, which
  replaces `DefaultMainPass` with its own main pass, so the patched class was not the one in
  use.

So the test that "verified" this exercised neither the clear-strip nor the `LOAD`. The
pause-menu flash improvement I signed off on most likely came from the null-renderPass guard
in the other commit, not from this change. I extrapolated from reading `dev` and assuming
1.21.1 matched, and the in-game evidence I leaned on was contaminated.

The crash guard is unaffected by any of this. Rather than close this PR and scatter the
history across a new number, I have force-pushed it to carry only that guard, and retitled
and rewritten the body to match. The flash commit is gone from the branch; this PR is now
just the crash fix.

Sorry for the round trip.

Chaos02:
Separate finding from the same tracing session, flagging it here since Beryl has no issue
tracker and this is relevant to any main-pass-level change in VulkanMod.

Beryl (0.2.0-alpha) replaces `DefaultMainPass` with `net.beryl.render.ShaderMainPass`, which
issues its own full-screen `DEPTH|COLOR` clear every frame from inside `beginFrame`. Traced
call, present on every frame:

```
Renderer.clearAttachments mask=0x4100 clearColor=[0.1176, 0.1176, 0.1176, 1.0]
	caller=net.beryl.render.ShaderMainPass.begin:57
	<- Renderer.beginMainRenderPass:308
	<- Renderer.beginFrame:289
```

The practical consequence for review here: **any fix made at the `DefaultMainPass` level is
inert for Beryl users**, because `DefaultMainPass` is not the class in use for them. That is
what invalidated my own verification of this PR, and it may be worth keeping in mind for
main-pass changes generally.

One related data point, since it points at Beryl rather than VulkanMod: A/B-ing Beryl off
makes the launch/menu flash I was originally chasing disappear.

Beryl is closed-source so I only have stack frames here, and I am not speculating past them.

